PR #46, opened 2021-09-30
Specify issues for todos

mike0sv:
Signed-off-by: mike0sv <mike0sv@gmail.com>


aguschin (approved):
It's great you've decided to create issues for those todos. I've added some more where it made sense.

PR #69, opened 2021-10-11
fix github ls

### LINKED ISSUE #64: Can't list models in the remote Git repo
One should be able to list models in remote Git repo. For example,
```
mlem ls https://github.com/iterative/example-mlem
mlem ls models https://github.com/iterative/example-mlem
```
or
```
mlem ls --repo https://github.com/iterative/example-mlem
mlem ls models --repo https://github.com/iterative/example-mlem
```

mike0sv:
Closes #64

### DIFF
--- a/mlem/cli/info.py
+++ b/mlem/cli/info.py
@@ -6,6 +6,7 @@ import click
 from fsspec.implementations.local import LocalFileSystem
 
 from mlem.cli.main import cli
+from mlem.core.meta_io import get_fs
 from mlem.core.metadata import load_meta
 from mlem.core.objects import (
     MLEM_DIR,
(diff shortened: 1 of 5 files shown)


aguschin (changes requested):
No comments about the code, but `mlem ls all --repo https://github.com/iterative/example-mlem` gives empty output. I've updated the test, so now it takes this into account (and fails).

aguschin (approved):
@mike0sv, in some future we will get requests about on-prem gitlab, for example. Will we be able to tell the difference between https://some-domain.com/some-path being Git repo or just https FS? If not, this may lead to some confusion for users here.

PR #77, opened 2021-10-13
Fix load from dvc

mike0sv:


aguschin (approved):
One thing to notice: right now `need_example_auth` effectively disables bunch of checks, because in those checks we load/get `data/model` only. And `data/model` now is stored in dvc in both `main` and `store-with-dvc` branches. It's better to keep those two separated, so when we are finished with updating `example-mlem`, I'll get back to tests and review those usages.

PR #89, opened 2021-10-21
ls command

### LINKED ISSUE #14: Design `mlem.api.list()` for common model registry case
Use Model Registry in Studio as an example of how people would like to see Model registry => This may affect `mlem.api.list()`

Beside listing **all** models in a repo, we may want to list all versions of one model.

As a corner case, it can be a monorepo with `.mlem` located not in repo root, but in a subfolder. We need to add option to list models in such cases.

mike0sv:
closes #14

Signed-off-by: mike0sv <mike0sv@gmail.com>

### DIFF
--- a/.github/workflows/check-test-release.yml
+++ b/.github/workflows/check-test-release.yml
@@ -32,7 +32,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python: [3.6, 3.7, 3.8, 3.9]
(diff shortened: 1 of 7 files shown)


aguschin (approved):

aguschin (approved):
Had few questions, but nothing major.

aguschin (approved):
🔥 Let's merge this! 🚒

PR #97, opened 2021-10-26
Add heroku functionality

### LINKED ISSUE #96: Add deployment to heroku
MLEM is designed to be a deployment tool: once you [saved your model](https://www.notion.so/iterative/Tutorial-9089de99f9da4d1f8926f900c797208e), you should be able to deploy it to heroku, sagemaker or any other provider. To support those, MLEM relies on extensions (plugins), which create concrete implementation for abstract classes from [MLEM's core](https://github.com/iterative/mlem/tree/main/mlem/core). Plugins also extend MLEM's other capabilities: they add ML frameworks and data libraries support. To see the latter, check out modules [in this folder](https://github.com/iterative/mlem/tree/main/mlem/contrib). Upon MLEM's import plugins [are registered](https://github.com/iterative/mlem/blob/main/mlem/__init__.py#L15).

In this task, we need to add plugin for deployment on heroku. We already had a draft on that functionality (it's even was working at some moment), now we need to:
1. fix it and make it work again, and optionally improve it (or do it from scratch 🔨 )
2. Beside that, it would be good to have a concise report, describing major changes and decisions made in the process, with one or two supporting sentences clarifying why it was done the way it is. 
3. If there were decent alternatives to decisions made, let's mention them also.

If the task turn out to be too hard, let's do this step by step: fix part of functionality and then discuss the progress.

PR with deployment functionality, which apparently doesn't work: https://github.com/iterative/mlem/pull/97

To give usage example, deployment demo we had previously was executed by the following script (note, that as the first step it would be enough to implement `mlem env create staging heroku` and `mlem deploy data/model staging`, or even only one of them):

```shell
if [[ -z "${HEROKU_API_KEY}" ]]; then
  echo please set HEROKU_API_KEY var
  exit 1
fi
REGISTRY_HEROKU_COM_PASSWORD=${HEROKU_API_KEY}
REGISTRY_HEROKU_COM_USERNAME=_

echo creating heroku staging env

mlem env create staging heroku

echo Deploying

mlem deploy data/model staging

echo Waiting for start up

DEPLOYED=false
for i in {1..15}; do {
  if [[ $(mlem status staging data/model) != 'up' ]]; then {
    echo .
    sleep 2;
    }
    else
    {
      echo Done
      DEPLOYED=true
      break
    }
  fi
  };
done

if [[ "${DEPLOYED}" != 'true' ]]; then {
  echo Deploy not completed, run 'mlem status staging data/model'
  exit 1
  }
fi

# echo Requesting remote
# mlem apply-remote -m predict_proba data/model data/remote_pred data/test_x

echo dont forget to run 'mlem destroy staging data/model'```

aguschin:
Closes #96


aguschin (commented):
Cool, @mike0sv! Looks like deployment is finally working (I didn't check myself yet though 😂 ).

madhur-tandon (approved):

aguschin (commented):
Cool! Let's ship it!

PR #131, opened 2021-12-10
art sizes and hashes

### LINKED ISSUE #90: Add hashes and size of artifacts to meta
If we store hashes, then meta `mlem.yaml` will change when artifacts are updated. This is the case, for example, when you change model hyperparameters. It seem to be good idea if meta would also change.

On the second thought, let's also add artifact size.

mike0sv:
closes #90


aguschin (approved):
Looks good 👍 
Would be good if you add few words here about why you went with md5 and not some other, considering that issue you posted ^

PR #134, opened 2021-12-10
import api and cli WIP

mike0sv:


aguschin (commented):
Cool functionality!
Looks good as a first draft.
Before settling on the command name, let's consider other alternatives to check that we don't miss a more suitable one. Maybe something from this? `wrap`, `codify`, `absorb`, `incorporate`, `capture`, `integrate`?

aguschin (changes requested):

aguschin (approved):

PR #145, opened 2022-01-11
add tests for fastapi

madhur-tandon:
Solves https://github.com/iterative/mlem/issues/42

@aguschin @mike0sv I have added some tests, I need help in two places which I have mentioned as comments.


aguschin (approved):
LGTM! 🔥  If @mike0sv doesn't have anything to add, let's merge!

PR #174, opened 2022-03-17
remove old artifacts

### LINKED ISSUE #88: Old, not relevant artifacts doesn't get removed
```python
from sklearn.datasets import load_iris
import mlem

train, target = load_iris(return_X_y=True)
mlem.api.save(train, "train")

train, target = load_iris(return_X_y=True, as_frame=True)
mlem.api.save(train, "train")
```

Then we have two artifacts:
```shell
train
├── artifacts
│   ├── data.csv
│   └── data.npz
└── mlem.yaml
```

mike0sv:
closes #88

### DIFF
--- a/mlem/core/artifacts.py
+++ b/mlem/core/artifacts.py
@@ -80,6 +80,10 @@ class Artifact(MlemObject, ABC):
         raise NotImplementedError
 
     @abstractmethod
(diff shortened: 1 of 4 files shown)


aguschin (approved):
Cool! :shipit:

PR #180, opened 2022-03-23
fix xfailing s3 tests

### LINKED ISSUE #110: Failing s3 tests on CI
Strange behavior, inconsistent failing of tests with s3 remote.
The actual exception is raised in `find_mlem_root` because fs.exists does not work and `_path` var gets stripped to just `"s3:"`, which is invalid s3 path because it lacks bucket. 
One can investigate different runs of this PR https://github.com/iterative/mlem/pull/105
Commits with inconsistent runs https://github.com/iterative/mlem/actions/runs/1436960627
https://github.com/iterative/mlem/actions/runs/1437008033

My guess that it depends on github actions worker (maybe some strange caching is involved).
Also, commit with unpinned versions for s3 packages is always failing
https://github.com/iterative/mlem/actions/runs/1436980743

mike0sv:
closes #110

### DIFF
--- a/tests/api/test_commands.py
+++ b/tests/api/test_commands.py
@@ -16,7 +16,7 @@ from mlem.core.meta_io import MLEM_DIR, MLEM_EXT
 from mlem.core.metadata import load
 from mlem.core.objects import DatasetMeta, MlemLink, ModelMeta
 from mlem.utils.path import make_posix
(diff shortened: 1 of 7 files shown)


aguschin (approved):
Great you uncovered that one!
So, the problem was in conflicting paths from different GH Action jobs?

PR #181, opened 2022-03-24
Feature/new apply

### LINKED ISSUE #25: Disambiguate passing multiple datasets to `apply`
@mike0sv comment:
> it seems that this logic will apply model to multiple datasets in one run. My vision was to apply model once, but using multiple datasets (for example, .fit with X and y)
  
It seems we need to support both approaches. I had a case in mind when you have a folder with images and want to run your NN on all of them at once.

### LINKED ISSUE #132: Can't save `apply` results outside of .mlem/ dir
```
mlem apply latest data/test_x -m predict_proba -o ./data/pred
```
dumps result under `.mlem/dataset/data/pred`. I'm pretty sure there are cases when a user would expect it to see it in just `./data/pred`. Let's do something about this, at least let's add `--external` param so you could control this?

mike0sv:
closes #25 
closes #132

### DIFF
--- a/mlem/api/commands.py
+++ b/mlem/api/commands.py
@@ -60,9 +60,6 @@ def apply(
         If `output=None`, returns results for given data.
             Otherwise returns None.
 
-    # TODO https://github.com/iterative/mlem/issues/25
-    # one may want to pass several objects instead of one as `data`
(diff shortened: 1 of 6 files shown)


aguschin (approved):
Yep! Let's make it simpler.

PR #183, opened 2022-04-02
remove serialize and deserialize from DatasetType

### LINKED ISSUE #157: DatasetType and DatasetSerializer both have overlapping methods
The functions `serialize` and `deserialize` are present in both `DatasetType` and `DatasetSerializer` classes.

But by convention, they should only be present in the `DatasetSerializer` class.
Having them present as a part of both means that one could get away with only inheriting from `DatasetType` rendering the presence of `DatasetSerializer` redundant and useless.

An example of such a class would be `NumpyNumberType` in `mlem/contrib/numpy.py` which implements methods such as `serialize` by virtue of being inherited from `DatasetType`. This defeats the purpose of having a separate `DatasetSerializer` class altogether.

CC: @mike0sv

madhur-tandon:
Fixes https://github.com/iterative/mlem/issues/157

### DIFF
--- a/.github/workflows/check-test-release.yml
+++ b/.github/workflows/check-test-release.yml
@@ -74,6 +74,8 @@ jobs:
       run: |
         pip install --upgrade pip setuptools wheel
         pip install pre-commit .[tests] --use-deprecated=legacy-resolver
+    - run: pre-commit run pylint -a -v --show-diff-on-failure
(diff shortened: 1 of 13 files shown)


mike0sv (commented):
👍

mike0sv (approved):

PR #196, opened 2022-04-14
add external to apply and default requirement type

mike0sv:


aguschin (approved):
:shipit: :shipit: :shipit:

PR #206, opened 2022-04-24
WIP: Support batch-reading for data-types with chunksize parameter

### LINKED ISSUE #23: Interface for batch loading for datasets
For big datasets and potentially for multi-file datasets (eg collection of images) we need an alternative way of reading data in batches. Potential use-case:
`mlem apply model some-very-big.csv --batch 10000` 
1. Design interface changes. Probably this should be a new method in `DatasetReader`, also `Dataset` and `DatasetMeta` could be affected. For some dataset types batching is impossible, so we should either ignore batch option and fallback to regular read, or raise an error (this should be configurable). Also, support for batch reading should be specified for each dataset type.
2. Add new option to `mlem apply` cli and change underlying API call to support this feature
3. Implement batch reading for pandas as POC, other dataset types may not support this

terryyylim:
### Context
Some datasets are large and rather than dealing them with one big block, we could split the data into chunks. This PR adds batch-reading support for data formats which provides the `chunksize` parameter using the Pandas API.

Supported formats:

- [x] CSV
- [x] JSON
- [x] Stata

### Modifications
- `mlem/cli/apply.py` - Added `batch` parameter when calling apply method - supports both importing data on/off-the-fly workflows
- `mlem/api/utils.py` - Added batch reading support when getting Dataset value
- `mlem/core/errors.py` - Added new errors `UnsupportedDatasetBatchLoadingType` and `DatasetBatchLoadingJSONError` for batch-reading workflows
- `mlem/contrib/pandas.py` - Added batch-reading support for CSV, JSON data formats
- `tests/contrib/test_pandas.py` - Added tests for supported batch-reading data formats, exception tests for unsupported batch-reading data formats 

Fixes https://github.com/iterative/mlem/issues/23

### DIFF
--- a/mlem/api/commands.py
+++ b/mlem/api/commands.py
@@ -58,6 +58,7 @@ def apply(
     output: str = None,
     link: bool = None,
     external: bool = None,
+    batch: Optional[int] = None,
 ) -> Optional[Any]:
     """Apply provided model against provided data
 
@@ -85,7 +86,7 @@ def apply(
         resolved_method = PREDICT_METHOD_NAME
     echo(EMOJI_APPLY + f"Applying `{resolved_method}` method...")
     res = [
-        w.call_method(resolved_method, get_dataset_value(part))
+        w.call_method(resolved_method, get_dataset_value(part, batch))
         for part in data
     ]
     if output is None:
@@ -412,6 +413,7 @@ def import_object(
     target_repo: Optional[str] = None,
     target_fs: Optional[AbstractFileSystem] = None,
     type_: Optional[str] = None,
+    batch: Optional[int] = None,
     copy_data: bool = True,
     external: bool = None,
     link: bool = None,
@@ -426,7 +428,7 @@ def import_object(
         if type_ not in ImportHook.__type_map__:
             raise ValueError(f"Unknown import type {type_}")
         meta = ImportHook.__type_map__[type_].process(
-            loc, copy_data=copy_data, modifier=modifier
+            loc, copy_data=copy_data, modifier=modifier, batch=batch
         )
     else:
         meta = ImportAnalyzer.analyze(loc, copy_data=copy_data)
--- a/mlem/cli/apply.py
+++ b/mlem/cli/apply.py
@@ -60,6 +60,9 @@ def apply(
         # TODO: change ImportHook to MlemObject to support ext machinery
         help=f"Specify how to read data file for import. Available types: {list_implementations(ImportHook)}",
     ),
+    batch: Optional[int] = Option(
+        None, "-b", "--batch", help="Batch size for reading data in batches."
+    ),
     link: bool = option_link,
     external: bool = option_external,
     json: bool = option_json,
@@ -83,7 +86,11 @@ def apply(
     with set_echo(None if json else ...):
         if import_:
             dataset = import_object(
-                data, repo=data_repo, rev=data_rev, type_=import_type
+                data,
+                repo=data_repo,
+                rev=data_rev,
+                type_=import_type,
+                batch=batch,
             )
         else:
             dataset = load_meta(
@@ -92,6 +99,7 @@ def apply(
                 data_rev,
                 load_value=True,
                 force_type=DatasetMeta,
+                batch=batch,
             )
         meta = load_meta(model, repo, rev, force_type=ModelMeta)
 
@@ -102,6 +110,7 @@ def apply(
             output=output,
             link=link,
             external=external,
+            batch=batch,
         )
     if output is None and json:
         print(
--- a/mlem/core/dataset_type.py
+++ b/mlem/core/dataset_type.py
@@ -404,6 +404,10 @@ class DatasetReader(MlemObject, ABC):
     def read(self, artifacts: Artifacts) -> DatasetType:
         raise NotImplementedError
 
+    @abstractmethod
+    def read_batch(self, artifacts: Artifacts, batch: int) -> DatasetType:
+        raise NotImplementedError
+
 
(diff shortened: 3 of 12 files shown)


terryyylim (commented):
Left some comments to explain what I changed, hopefully it helps with the review!

mike0sv (commented):
I commented in the code a bit, but main idea is following: you implemented `batch` as a flag for regular read. But actually we need like a separate type of loading data. 
It should return some lazy iterable object that would yield batches, preferably of the same type that the whole dataset is. For example, `apply` flow should change something like that:
```
if batch is None:
  # existing logic
  res = apply(model, dataset.get_value())
else:
  res = [apply(model, batch) for batch in dataset.iter_batches(batch)]
```

You don't need to worry about batch writing or merging different batches for now

terryyylim:
@mike0sv I think my forked repo is no longer synced with this repository because it's made public now (my commits are no longer updated here). Shall I make a new PR?

tests/conftest.py
```diff
@@ -328,6 +329,43 @@ def dataset_write_read_check(
...
+        df_iterable: Iterator = reader.read_batch(artifacts, BATCH_SIZE)
+        df = None
+        col_types = None
+        while True:
+            try:
+                chunk = next(df_iterable)
+                if df is None:
+                    df = pd.DataFrame(columns=chunk.columns, dtype=col_types)
+                    col_types = {
+                        chunk.columns[idx]: chunk.dtypes[idx]
```
mike0sv, line 358: `zip` would fit nicely here

mlem/core/dataset_type.py
```diff
@@ -404,6 +414,10 @@ class Config:
     def read(self, artifacts: Artifacts) -> DatasetType:
         raise NotImplementedError
 
+    @abstractmethod
+    def read_batch(self, artifacts: Artifacts, batch: int) -> Iterator:
```
mike0sv, line 418: This should actually be `Iterator[DatasetType]`

mlem/core/metadata.py
```diff
@@ -142,6 +145,7 @@ def load_meta(
     follow_links: bool = True,
     load_value: bool = False,
     fs: Optional[AbstractFileSystem] = None,
+    batch: Optional[int] = None,
```
mike0sv, line 148: `load_meta` should not have batch arg: you are only loading metadata. `load_value` here is for convenience. If you need batching, you should set `load_value=False` and then call `read_batch` on `DatasetMeta` directly

mlem/core/objects.py
```diff
@@ -668,6 +668,9 @@ def write_value(self) -> Artifacts:
     def load_value(self):
         self.dataset = self.reader.read(self.relative_artifacts)
 
+    def load_batch_value(self, batch: int):
```
mike0sv, line 671: this should be changed to something like `read_batch`, return `Iterator[DatasetType]` and not set any values. `self.dataset` field is for whole dataset value, and if you wish to use batching, that means you don't want to load whole dataset into memory. So, `read_batch` will return lazy iterator that you can iterate on and get `DatasetType`s to work with. For now lets say that those `DatasetType`s should be the same as the `reader` is holding (==`DatasetType` of the whole dataset).

mlem/cli/apply.py
```diff
@@ -83,7 +86,11 @@ def apply(
     with set_echo(None if json else ...):
         if import_:
             dataset = import_object(
-                data, repo=data_repo, rev=data_rev, type_=import_type
+                data,
+                repo=data_repo,
+                rev=data_rev,
+                type_=import_type,
+                batch=batch,
```
mike0sv, line 93: for now let's ignore this branch and focus on the other one without import. once we done, we can discuss how to approach this (I am not sure myself)

mlem/cli/apply.py
```diff
@@ -92,6 +99,7 @@ def apply(
                 data_rev,
                 load_value=True,
```
mike0sv, line 100: as per comments in `metadata.py` and `objects.py`, this should be `load_meta(..., load_value=batch is None)`

mlem/api/commands.py
```diff
@@ -85,7 +86,7 @@ def apply(
         resolved_method = PREDICT_METHOD_NAME
     echo(EMOJI_APPLY + f"Applying `{resolved_method}` method...")
     res = [
-        w.call_method(resolved_method, get_dataset_value(part))
+        w.call_method(resolved_method, get_dataset_value(part, batch))
```
mike0sv, line 89: Here you should flatten all the batches if there are any. Here the example of how I see it:
Suppose you have `dataframe=[1,2,3,4]` (I mean 1 column, 4 rows) saved to csv file. You load its metadata without loading the value, let's say `dt = DatasetMeta(dataset_type=DataFrameType(...), ...)`. And you call `apply` with data=[dt]. If batch arg is not provided, what will happen is `get_dataset_value` will load the actual dataframe and in the end `res = [w.call_method(..., dataframe([1,2,3,4))]`. But if you provided `batch=2`, `dt. read_batch` should be called. If you iterate through it, you will get 2 parts of the dataframe, and in the end `res=[w.call_method(..., dataframe([1,2])), w.call_method(..., dataframe([3,4]]))]`

terryyylim: 👍🏻 Makes sense, I implemented something similar yesterday, but I couldn't sync the changes to this PR because the private fork no longer points to this repository.

terryyylim:
Closing the PR because private forks can't sync to this repository anymore since it went public. I've pushed the changes to a separate PR (https://github.com/iterative/mlem/pull/216) after forking the public repository.

PR #216, opened 2022-04-29
Support batch-reading for data-types with chunksize parameter

### LINKED ISSUE #23: Interface for batch loading for datasets

terryyylim:
### Context
Some datasets are large and rather than dealing them with one big block, we could split the data into chunks. This PR adds batch-reading support for data formats which provides the `chunksize` parameter using the Pandas API.

Related PR: https://github.com/iterative/mlem/pull/206

Supported formats:

- [x] CSV
- [x] JSON
- [x] Stata

### Modifications
- `mlem/cli/apply.py` - Added `batch` parameter when calling apply method - supports both importing data on/off-the-fly workflows
- `mlem/api/utils.py` - Added batch reading support when getting Dataset value
- `mlem/core/errors.py` - Added new errors `UnsupportedDatasetBatchLoadingType` and `UnsupportedDatasetBatchLoading` for batch-reading workflows
- `mlem/contrib/pandas.py` - Added batch-reading support for CSV, JSON, Stata data formats
- `tests/contrib/test_pandas.py` - Added tests for supported batch-reading data formats

Fixes https://github.com/iterative/mlem/issues/23

### DIFF
--- a/mlem/api/commands.py
+++ b/mlem/api/commands.py
@@ -90,13 +90,9 @@ def apply(
         res: Any = []
         for part in data:
             batch_dataset = get_dataset_value(part, batch)
-            while True:
-                try:
-                    chunk = next(batch_dataset)
-                    preds = w.call_method(resolved_method, chunk.data)
-                    res = [*res, *preds]
-                except StopIteration:
-                    break
+            for chunk in batch_dataset:
+                preds = w.call_method(resolved_method, chunk.data)
+                res += [*preds]
         res = [np.array(res)]
     else:
         res = [
--- a/mlem/api/commands.py
+++ b/mlem/api/commands.py
@@ -5,6 +5,7 @@ import posixpath
 from collections import defaultdict
 from typing import Any, Dict, Iterable, List, Optional, Type, Union
 
+import numpy as np
 from fsspec import AbstractFileSystem
 from fsspec.implementations.local import LocalFileSystem
 
@@ -58,6 +59,7 @@ def apply(
     output: str = None,
     link: bool = None,
     external: bool = None,
+    batch: Optional[int] = None,
 ) -> Optional[Any]:
     """Apply provided model against provided data
 
@@ -84,10 +86,23 @@ def apply(
     except WrongMethodError:
         resolved_method = PREDICT_METHOD_NAME
     echo(EMOJI_APPLY + f"Applying `{resolved_method}` method...")
-    res = [
-        w.call_method(resolved_method, get_dataset_value(part))
-        for part in data
-    ]
+    if batch:
+        res: Any = []
+        for part in data:
+            batch_dataset = get_dataset_value(part, batch)
+            while True:
+                try:
+                    chunk = next(batch_dataset)
+                    preds = w.call_method(resolved_method, chunk.data)
+                    res = [*res, *preds]
+                except StopIteration:
+                    break
+        res = [np.array(res)]
+    else:
+        res = [
+            w.call_method(resolved_method, get_dataset_value(part))
+            for part in data
+        ]
     if output is None:
         if len(res) == 1:
             return res[0]
--- a/mlem/core/objects.py
+++ b/mlem/core/objects.py
@@ -14,6 +14,7 @@ from typing import (
     Dict,
     Generic,
     Iterable,
+    Iterator,
     List,
     Optional,
     Tuple,
@@ -672,6 +673,9 @@ class DatasetMeta(_WithArtifacts):
 
     def load_value(self):
         self.dataset = self.reader.read(self.relative_artifacts)
+
+    def read_batch(self, batch: int) -> Iterator[DatasetType]:
+        return self.reader.read_batch(self.relative_artifacts, batch)  # type: ignore
 
     def get_value(self):
         return self.data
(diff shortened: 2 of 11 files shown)


mlem/api/commands.py
```diff
@@ -84,10 +86,23 @@ def apply(
...
     echo(EMOJI_APPLY + f"Applying `{resolved_method}` method...")
-    res = [
-        w.call_method(resolved_method, get_dataset_value(part))
-        for part in data
-    ]
+    if batch:
+        res: Any = []
+        for part in data:
+            batch_dataset = get_dataset_value(part, batch)
+            while True:
```
mike0sv, line 93: why not `for chunk in batch_dataset:`?

mlem/api/commands.py
```diff
@@ -84,10 +86,23 @@ def apply(
...
+                try:
+                    chunk = next(batch_dataset)
+                    preds = w.call_method(resolved_method, chunk.data)
+                    res = [*res, *preds]
```
mike0sv, line 97: `+=` will do the job :)

mlem/core/objects.py
```diff
@@ -673,6 +674,9 @@ def write_value(self) -> Artifacts:
     def load_value(self):
         self.dataset = self.reader.read(self.relative_artifacts)
 
+    def read_batch(self, batch: int) -> Iterator[DatasetType]:
+        return self.reader.read_batch(self.relative_artifacts, batch)  # type: ignore
```
mike0sv, line 678: why `type: ignore`?

tests/conftest.py
```diff
@@ -328,6 +329,40 @@ def dataset_write_read_check(
                 assert new.data == dataset.data
 
 
+def dataset_write_read_batch_check(
```
mike0sv, line 332: It seems this one can be used only for `pandas` datasettypes, so it should go to `test_pandas` or made more general (not sure how)

tests/contrib/test_pandas.py
```diff
@@ -120,6 +124,24 @@ def test_simple_df(data, format):
     )
 
 
+@for_all_formats(
```
mike0sv, line 127: probably easier to list included formats :)

mlem/core/objects.py
terryyylim: Ah, so `mypy` was complaining about this previously and I've circumvented this by an assertion check rather than ignoring. Will remove `type: ignore`.

```
mlem/core/objects.py: note: In member "read_batch" of class "DatasetMeta":
mlem/core/objects.py:685:16: error: Item "None" of "Optional[DatasetReader]"
has no attribute "read_batch"  [union-attr]
```

terryyylim:
@mike0sv , some thoughts on the follow workflows:

**Batching on import (Import data-on-the-fly) workflow**
Currently, we call `mlem/api/commands.py (import_object method)` which calls ImportHook class object’s process method
This call loads the entire dataset into memory via `mlem/core/metadata.py’s (get_object_metadata method)` and binds it to dataset attribute.

I think what we can do for batch reading, would be to:
1. Add a new `batch` argument to process method (https://github.com/iterative/mlem/blob/main/mlem/contrib/pandas.py#L535)
2. Call retrieval of object metadata:
     a. If `batch` argument is not provided - Call retrieval of object metadata via get_object_metadata method (https://github.com/iterative/mlem/blob/main/mlem/core/metadata.py#L25)
     b. If `batch` argument is provided, retrieve only metadata of the dataset, without binding it to dataset attribute, i.e populating reader_cache information and setting dataset attribute to none
Something like: 
```
reader_cache={'dataset_type': {'columns': ['', 'sepal length (cm)', 'sepal width (cm)', 'petal length (cm)', 'petal width (cm)'], 'dtypes': 				['int64', 'float64', 'float64', 'float64', 'float64'], 'index_cols': [''], 'type': 'dataframe'}, 'format': 'csv', 'type': 'pandas'} dataset=None
```
By doing so, when apply method (https://github.com/iterative/mlem/blob/main/mlem/cli/apply.py#L98) is called subsequently, it’ll trigger the batch-reading workflow as implemented in the current PR (https://github.com/iterative/mlem/pull/216/files#diff-7c25a5730147dd254f2b720216ef3c356a9828b85afdbea0c88123a85212bd18R17).

**Batch writing**
For batch writing, this concerns the `https://github.com/iterative/mlem/blob/main/mlem/core/metadata.py#L47 (save method)`, which calls the dump method and triggers writing via `https://github.com/iterative/mlem/blob/main/mlem/core/objects.py#L486 (get_artifacts method)`.

We could expose a new `write_batch` method that considers 2 scenarios here:
1. Writing data in batches via _a new argument `batch` (int)_ passed to related function calls much like how batch-reading is done now, that does something like:
```
df.to_csv("path/to/save/file.csv", chunksize=batch)
```
2. AND writing data that is suitable for batch-reading via _a new argument `batch_read` (boolean)_, where internally within our eg.`contrib.pandas` module, we could utilise PANDAS_FORMATS's write_args that has been configured for batching in this PR i.e https://github.com/iterative/mlem/pull/216/files#diff-706e38d0ea328a5aa14720511896d8e86f7e75edbe79156defa0dea248bfd858R499.

PR #223, opened 2022-05-05
add checking python env

mike0sv:


aguschin (approved):
Wow, nice command. Do you think it's better to print both versions (required, existing) for debugging purposes? Now it's
```
$ mlem checkenv https://github.com/iterative/example-mlem-get-started/rf
⏳️ Loading meta from https://github.com/iterative/example-mlem-get-started/tree/main/.mlem/model/rf.mlem
❌ Wrong requirements: Wrong versions for ['pandas==1.4.2'].
```

and could be 

```
$ mlem checkenv https://github.com/iterative/example-mlem-get-started/rf
⏳️ Loading meta from https://github.com/iterative/example-mlem-get-started/tree/main/.mlem/model/rf.mlem
❌ Wrong requirements: `pandas==1.4.2` is required, but `==1.4.1` is installed
```

Issue #235, opened 2022-05-07
`mlem.save` don't find any requirements


aguschin:
Two problems I see in the code/output below:
1. neither `emoji` nor `numpy` is added to `.mlem` file
2. there is some `data.pkl` in `.mlem` file, which is confusing to me as a reader. Also, looks like another 🐞 

```python
import emoji
import numpy as np


def translate(text: str):
    """
    Translate dog barks to emoji, as you hear them
    """
    return " ".join(
        np.random.choice(list(emoji.EMOJI_DATA.keys()))
        for _ in text.split()
    )

print(translate("Woof woof!"))


import mlem

mlem.api.save(translate, "dog-bark-translator", sample_data="Woof woof!")
```

```yaml
$ cat dog-bark-translator.mlem
artifacts:
  data.pkl:
    hash: dfbceeaa2ee4f09adae5f33c9f0ae022
    size: 595
    uri: dog-bark-translator
model_type:
  io:
    type: pickle
  methods:
    __call__:
      args:
      - name: data
        type_:
          ptype: str
          type: primitive
      name: translate
      returns:
        ptype: str
        type: primitive
    predict:
      args:
      - name: data
        type_:
          ptype: str
          type: primitive
      name: predict
      returns:
        ptype: str
        type: primitive
  type: callable
object_type: model
```

PR #243, opened 2022-05-11
add exceptions for __main__

mike0sv:
Probably fix for #235


aguschin (approved):
Looks like some magic, but it works! Thanks, @mike0sv! 🧙‍♂️

PR #264, opened 2022-05-22
add core section

mike0sv:
This PR changes MLEM config to be contained in a `core` section. This way the first layer of keys of the config file is all sections and not core config keys + sections for other configs. This also enables MLEM to validate config changes. To do that I also added extension-like entrypoint for config classes so you can validate even external configs


aguschin (approved):
👍

PR #272, opened 2022-05-27
add meta flag to load

mike0sv:


omesser:
Generally speaking I think we shouldn't have 2 ways to do the same thing. So either `load_meta()` OR `load(..., as_meta=True)` no need for both, unless some specific restrictions require this.

Some questions to give me background:
1. Would the user be interested in interacting with the `MlemObject` directly? what are the benefits?
2. Why is "MlemObject" is called "meta" in user facing flags/methods and not mlem_object (load_mlem_object, as_mlem_object)?
3. If there's an existing `load_meta()` - why isn't `load()` using it today (duplicated logic?)

mike0sv:
closing as outdated

PR #273, opened 2022-05-27
Shell requirements

### LINKED ISSUE #235: `mlem.save` don't find any requirements

mike0sv:
probably fixes #235 
but I also found some other cases that are not fixed and wrote tests for them. Will fix later


aguschin (approved):
Works! Awesome!

Issue #203, opened 2022-04-21
Reading config prior to executing anything


aguschin:
Wrong config files cause MLEM to crash even before you actually need them.
Steps to reproduce:
```
$ mkdir tmp
$ cd tmp
$ mlem init
$ echo WRONG_TExT > .mlem/config
$ mlem -h
...
could not find expected ':'
  in "/Users/aguschin/Git/iterative/mlem/tmp/.mlem/config.yaml", line 3, column 1
```

I would suggest that by default we should initialize config without reading config file (I mean, here: `CONFIG = MlemConfig()` https://github.com/iterative/mlem/blob/main/mlem/config.py#L116). For that we can have special `NoFileConfig(BaseSettings)` class.

Then when user actually executes something on specific repo, we should use `MLEMConfig(NoFileConfig)` class.

I did the same in GTO because it makes sense there, not so sure about MLEM. WDYT @mike0sv ?

aguschin:
Now it works! Closing)

PR #276, opened 2022-05-30
Dynamic config loading

mike0sv:
In case config was updated after python process started, the changes would have no effect since config was loaded on startup. This was fixed for cases when config is used not on startup. Also now target project config is used where relevant


shcheklein (approved):

aguschin (approved):
Does this fix #203 btw?

PR #287, opened 2022-06-07
Refactor github and gitlab resolvers, implement bitbucket

### LINKED ISSUE #192: gitlab & bitbucket integration
Right now we can use github URLs to reference different objects, like `https://github.com/org/repo/path`
We need to add the same functionality for GL and BB. The easiest way is probably to implement a subset of fsspec AFS and new corresponding UriResolver the same way as for github

mike0sv:
closes #192


aguschin (approved):
Awesome!

PR #290, opened 2022-06-08
Support combining batched prediction results on DatasetType

### LINKED ISSUE #257: Add `numpy` or refactor this
#221 introduced numpy dependency in `api.py`: https://github.com/iterative/mlem/blob/main/mlem/api/commands.py#L7
We either need to remove it, or we need to add `numpy` to install_requires in `setup.py`.

terryyylim:
### Context
A bug was introduced in https://github.com/iterative/mlem/pull/221, where approach to combining prediction results incorrectly assumed type to be numpy array. This PR utilizes `DatasetAnalyzer` to first determine the type of the results prior to combining them, since the same model can output different data types.

### Modifications
- `mlem/api/commands.py` - Interpret data type prior to combining prediction results
- `mlem/contrib/numpy.py` - Add support for combining batch reading prediction results for nd.array

Fixes https://github.com/iterative/mlem/issues/257


aguschin:
I've updated the branch, now auth issues are gone. But there are some others I didn't investigate. @terryyylim do you mind taking a look?

PR #304, opened 2022-06-20
dynamic types support array, dict, ndarray

### LINKED ISSUE #299: support for dynamically shaped/sized symbolic data types
Currently data types such as DictType, NumpyNdarrayType do not support dynamic size or shape. The types are bound to static data sources. This is problematic when inference output type size/shape can vary based on input size/shape. The onnx supports dynamically shaped/sized symbolic shapes. Similar kind of datatypes should also be supported in MLEM. The ONNX majorly supports NDArray tensor, sequence and map types. NDArray shape can have unknown or named values. Sequence and map does not have size attribute.  
Changes needed:
- NDArray equivalent NumpyNdarrayType can be modified to support unknown/dynamic values in its shape. Named values support can be added in future if need be. Currently first dimension is treated as unknown.
- Sequence equivalent ArrayType supports dynamic size. The size needs to be -1.
- Map equivalent DictType is hard to modify to support dynamic size because in current implementation the dataset is tightly bound with the DictType. The keys of dataset are part of DictType. Need to add new datatype for Dict.  Need to see how often maps with dynamic set of keys are used in real world models.

maheshambule:
Fixes https://github.com/iterative/mlem/issues/299


mlem/contrib/numpy.py
```diff
@@ -148,10 +157,20 @@ def serialize(self, instance: np.ndarray):
         return instance.tolist()
 
     def _check_shape(self, array, exc_type):
-        if tuple(array.shape)[1:] != self.shape[1:]:
-            raise exc_type(
-                f"given array is of shape: {(None,) + tuple(array.shape)[1:]}, expected: {self.shape}"
+        if self.shape:
```
mike0sv, line 160: this will check if shape is None or shape is empty tuple. Maybe check if it is None?

mlem/core/data_type.py
```diff
@@ -449,44 +452,77 @@ def is_object_valid(cls, obj: Any) -> bool:
...
     type: ClassVar[str] = "dict"
     item_types: Dict[str, DataType]
 
     @classmethod
-    def is_object_valid(cls, obj: Any) -> bool:
-        return isinstance(obj, dict)
-
-    @classmethod
-    def process(cls, obj: Any, **kwargs) -> "DictType":
+    def analyze(cls, obj, **kwargs):
```
mike0sv, line 520: why override `analyze`?

mlem/core/data_type.py
```diff
@@ -577,6 +613,135 @@ def read_batch(
...
 
 
+class DynamicDictType(DataType, DataSerializer):
+    """
+    Dynamic DataType for dict without fixed set of keys
+    """
+
+    type: ClassVar[str] = "d_dict"
+
+    key_type: DataType
```
mike0sv, line 623: keys should be immutable. and json keys are always strings. Since you write it as json, how would you recreate arbitrary type?

tests/contrib/test_numpy.py
```diff
@@ -124,26 +282,10 @@ def test_ndarray_serialize_failure(nat, obj):
...
     with pytest.raises(DeserializationError):
-        nat.deserialize(obj)
+        nat[1].deserialize(obj)
 
 
 def test_requirements():
     assert get_object_requirements(
         NumpyNdarrayType(shape=(0,), dtype="int")
     ).modules == ["numpy"]
-
```
mike0sv, line 134: please leave the lisence in place :)

mike0sv (approved):

aguschin:
I re-run some failing tests locally, they pass. Must be auth/creds issue.

aguschin:
Ok, I think I found the reason of this auth issue @maheshambule. Let me try to rerun the tests.

aguschin:
Ok, I was pointed towards this issue https://github.com/iterative/cml/pull/574 by @0x2b3bfa0 
Will create a PR that does the same (should fix auth issue), then get back to this PR to re-run this.

aguschin (approved):

aguschin (approved):

aguschin:
Finally, this works! Thank you @maheshambule and @0x2b3bfa0. Merging after tests pass!

aguschin:
Looks like there are some issues unrelated to auth. @maheshambule, do you mind checking them out?

aguschin:
Great, finally we can merge this! @maheshambule, do you want me to go ahead or you will do it yourself?

aguschin:
Sorry! Thought you could do that 🤦‍♂️
Thanks for the great work!

Issue #149, opened 2022-01-15
Fix double deserialisation


madhur-tandon:
While writing tests for fastapi here: https://github.com/iterative/mlem/pull/145

It is observed that deserialisation happens at two places:

- pydantic's internals turning payload into `response_model`
- call to `serializers[a.name].deserialize()` inside `fastapi.py`

I wonder if removing `parse_obj_as(response_model, response)` for the `handler()` function in `fastapi.py` fixes this?
We ideally need to ensure deserialisation happens in a single pass and in correct manner.

aguschin:
closed by https://github.com/iterative/mlem/pull/308

PR #308, opened 2022-06-21
Add ignore-errors for ls CLI & API

mike0sv:
also better CLI output for ValidationError
fix get_model for primitives


mike0sv:
suddenly it also closes https://github.com/iterative/mlem/issues/149

aguschin (approved):

Issue #233, opened 2022-05-06
CLI: too many `-c` arguments without enough detail about what's required


dberenbaum:
Examples from the docs:

`$ mlem pack rf pip -c target=build/ -c package_name=example_mlem_get_started`

`$ mlem apply-remote http test_x.csv -c host="0.0.0.0" -c port=8080 --json`

`$ mlem create env heroku staging -c api_key=<you api key>`

`$ mlem create deployment heroku myservice -c app_name=example-mlem-get-started -c model=model -c env=staging`

`$ mlem apply-remote http test_x.csv -c host=http://my-mlem-service.herokuapp.com -c port=80 --json`

It feels like `-c` is being used as a catch-all and a way to avoid specifying the actual expected options for each command. It's not clear what each `-c` option is doing (if anything), which ones are required, and what values are expected.

mike0sv:
We have `mlem types` now, but we should try and add this to help

jorgeorpinel:
Where can we find a list of all possible config options per command? I was pointed to https://github.com/iterative/mlem/blob/main/tests/contrib/test_heroku.py before but it's unclear what the config options are in there.

mike0sv:
you can run `mlem types env heroku` and it will print all available options for it
in general, `mlem types <object type> <implementation name>`
The reason to use `-c` instead of `--<option name>` is because different implementations have different set of options. and adding options dynamically will probably require a lot of our cli framework butchering. we might do it later
From documentation point of view, we need a dedicated page for each implementation (not only target envs, but other base classes too). It's huge amount of work but it needs to be done at some point

jorgeorpinel:
> dedicated page for each implementation (not only target envs, but other base classes too)

Agree: https://github.com/iterative/mlem.ai/issues/143

Documenting all existing config options may also help us design a more segmented config CLI later (for this issue).

aguschin:
For the record, @mike0sv is testing a way to show all of these `-c` args in command's help in #363

aguschin:
closed by #363 🎸

PR #309, opened 2022-06-21
Docker container deployment 

mike0sv:


aguschin (approved):
LGTM!

One general feedback, the "declare and run" mechanics seems too involved for me each time I try to use it.

For a quicker "I run it and it worked" experience, I would suggest to consider cutting some corners:
1. generate docker image name automatically
2. using FastAPI server automatically
3. allowing to skip environment creation (don't dump it even)
4. allowing to skip deployment declaration creation (don't save it even)

Otherwise you need to read some docs... Which is boresome if you just want to check out how it works. Something like `$ mlem deploy run -m mymodel.mlem -e docker` would work I think.

WDYT? @mike0sv

mike0sv:
1. can be done
2. this will require fastapi dependencies, but might be done
3. hard. we might be able to automatically create env with all defaults, but we still need to dump it probably. or we can treat "no env" as default all the time
4. it can be done now (mlem deploy run can create it for you). not saving will prevent you from managing it after (no status, no teardown, no client requests) so you'll have to do all of this manually which is harder than just save declaration once

aguschin:
Ok, I feel like this should be a part of a bigger discussion, related to https://github.com/iterative/mlem/issues/233. This PR is good to be merged unless you want to add 1 and 2 here.

tests/contrib/test_docker/test_deploy.py
```diff
@@ -0,0 +1,116 @@
...
+    DockerContainerState,
+    DockerEnv,
+    DockerImage,
+)
+from mlem.contrib.fastapi import FastAPIServer
+from mlem.core.errors import DeploymentError
+from mlem.core.objects import DeployStatus
+from tests.contrib.test_docker.conftest import docker_test
+
+IMAGE_NAME = "mike0sv/ebaklya"
```
aguschin, line 20: Easter egg?

aguschin (commented):
🤔

tests/contrib/test_docker/test_deploy.py
mike0sv: kinda. it was used by tests before

Issue #326, opened 2022-06-29
Fix heroku test


aguschin:
This Heroku test is failing frequently. Especially when there are few CIs running in parallel. Rerunning it helps, but is very annoying. Need to investigate and fix.

```
FAILED tests/contrib/test_heroku.py::test_env_deploy_full
```

Examples:

1. "Cannot push docker image"
  https://github.com/iterative/mlem/runs/7092781603?check_suite_focus=true
  https://github.com/iterative/mlem/runs/7110470094?check_suite_focus=true
2. Assertion error, status is "starting", not "running"
  https://github.com/iterative/mlem/runs/7100085183?check_suite_focus=true
  https://github.com/iterative/mlem/runs/7109439795?check_suite_focus=true
3. "mlem.core.errors.DeploymentError: Name mlem-test-full-cycle-runner is already taken"
  https://github.com/iterative/mlem/runs/7110405418?check_suite_focus=true

UPD: this may benefit from solving holistically e.g. with [flaky](https://docs.pytest.org/en/7.1.x/explanation/flaky.html) or [retry](https://pypi.org/project/retry/).

aguschin:
Item 2 should be addressed by #328

PR #327, opened 2022-06-29
Bump Heroku tests limits

aguschin:
address #326


aguschin (commented):
Add with: python-version: '3.x'

PR #331, opened 2022-07-01
Mark some tests as flaky

### LINKED ISSUE #326: Fix heroku test

aguschin:
close #326 

See also https://github.com/dropbox/pytest-flakefinder for a way to find those tests

UPD: I've used `pytest-flakefinder` to run all tests 10 times, and only heroku test fails. From what I saw in CI, there are other tests that may fail, so I assume this is due to parallel runs. We can mark more tests as flaky once we find those.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -33,6 +33,7 @@ from mlem.runtime.interface import ModelInterface
 RESOURCES = "resources"
 
 long = pytest.mark.long
+flaky = pytest.mark.flaky
 MLEM_TEST_REPO_ORG = "iterative"
 MLEM_TEST_REPO_NAME = "mlem-test"
 MLEM_TEST_REPO = (
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -9,6 +9,7 @@ import numpy as np
 import pandas as pd
 import pytest
 from fastapi.testclient import TestClient
+from flaky import flaky  # noqa  # pylint: disable=unused-import
 from fsspec.implementations.local import LocalFileSystem
 from git import GitCommandError, Repo
 from requests import ConnectionError, HTTPError
@@ -33,7 +34,6 @@ from mlem.runtime.interface import ModelInterface
 RESOURCES = "resources"
 
 long = pytest.mark.long
-flaky = pytest.mark.flaky
 MLEM_TEST_REPO_ORG = "iterative"
 MLEM_TEST_REPO_NAME = "mlem-test"
 MLEM_TEST_REPO = (
--- a/tests/contrib/test_heroku.py
+++ b/tests/contrib/test_heroku.py
@@ -153,8 +153,10 @@ def _check_heroku_deployment(meta):
 
 
 def is_not_crash(err, *args):  # pylint: disable=unused-argument
-    time.sleep(10)
-    return not issubclass(err[0], DeploymentError)
+    needs_another_try = issubclass(err[0], DeploymentError)
+    if needs_another_try:
+        time.sleep(10)
+    return not needs_another_try
 
 
 @flaky(rerun_filter=is_not_crash, max_runs=2)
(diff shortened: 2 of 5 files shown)


tests/contrib/test_heroku.py
```diff
@@ -152,6 +152,7 @@ def _check_heroku_deployment(meta):
     assert len(res) == 1
 
 
+@flaky
```
shcheklein, line 155: please make sure that it retries by default and fails if no attempts are successful. We should not keep tests that we don't check results for.

aguschin: Checked, this works as expected.

tests/conftest.py
```diff
@@ -9,6 +9,7 @@
 import pandas as pd
 import pytest
 from fastapi.testclient import TestClient
+from flaky import flaky  # noqa  # pylint: disable=unused-import
```
mike0sv, line 12: why is this needed?

mike0sv:
for parallel runs we just need to use matrix values (os and python version) as prefixes

tests/conftest.py
aguschin: 1. We could use `pytest.mark.flaky`, but it doesn't support args like `max_runs`
2. `# noqa  # pylint: disable=unused-import` is needed so flake8 and pylint won't error out.

aguschin:
> for parallel runs we just need to use matrix values (os and python version) as prefixes

You already put `heroku_matrix = skip_matrix("ubuntu-latest", "3.7")`, so the problem not in parallel runs in a single CI, but in parallel runs in multiple CI. But we can append commit hexsha to Heroku deployment name for example, maybe that will solve something.

tests/contrib/test_heroku.py
```diff
@@ -152,6 +152,12 @@ def _check_heroku_deployment(meta):
     assert len(res) == 1
 
 
+def is_not_crash(err, *args):  # pylint: disable=unused-argument
+    time.sleep(10)
```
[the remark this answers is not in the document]
aguschin: To wait for some time before retrying. Just for confidence.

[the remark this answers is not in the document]
mike0sv: will it sleep even if test succeeded?

tests/conftest.py
mike0sv: Are we going to though? Is there any other implementation for flaky? :)
I did it for long because it's shorter :)

aguschin: Yes, there are - couple are listed here https://docs.pytest.org/en/7.1.x/explanation/flaky.html#plugins
I dunno if we are going to change this TBH, but see no harm in keeping it like it is right now.

tests/contrib/test_heroku.py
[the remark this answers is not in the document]
aguschin: No, but it will sleep if the test failed with another exception (not DeploymentError). Fixed this in the new commit.

mike0sv (approved):

PR #343, opened 2022-07-13
Enable `pip install mlem[dvc-s3]` for all relevant DVC dependencies

aguschin:
Added all relevant options from https://github.com/iterative/dvc/blob/main/setup.cfg#L74 (everything except `all`, `terraform`, `tests`).
Basically, extras for all [Storage types supported for remotes](https://dvc.org/doc/command-reference/remote/add#supported-storage-types).

The need for this was found out by @shcheklein [here](https://github.com/iterative/mlem.ai/pull/132#discussion_r919286747).

### DIFF
--- a/.github/workflows/check-test-release.yml
+++ b/.github/workflows/check-test-release.yml
@@ -85,12 +85,11 @@ jobs:
         brew install hdf5 wget c-blosc
         wget https://raw.githubusercontent.com/Homebrew/homebrew-core/fb8323f2b170bd4ae97e1bac9bf3e2983af3fdb0/Formula/libomp.rb
         brew install ./libomp.rb
-    # legacy-resolver makes install faster by order of magnitude
     - name: Install
       if: steps.cache.pip-cache-dir.cache-hit != 'true'
       run: |
         pip install --upgrade pip setuptools wheel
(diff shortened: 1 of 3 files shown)


mike0sv:
1. can we do this dynamically? like `extras = {..., **{f"dvc-{e}":[f"dvc[{e}]~=2.0"] for e in get_extras("dvc")}}`?
2. we don't need all of them for tests, so `extras["tests"] += [e for e in extras["all"] if not e.startswith("dvc-")]`

aguschin:
@mike0sv
1. Yes, we could, like this
```
import importlib_metadata

# add dvc extras
for e in importlib_metadata.metadata('dvc').get_all('Provides-Extra'):
    if e not in {"all", "terraform", "tests"}:
        extras[f"dvc-{e}"] = [f"dvc[{e}]~=2.0"]
```
But to use this, we need to first install `importlib_metadata`, which is not installed by default. Then we have a recursion: to run `setup.py`, you need to `pip install importlib_metadata`. But how to install `pip install importlib_metadata` first? Do you have some ideas?

2. Thanks, did that.

mike0sv (approved):

PR #355, opened 2022-07-14
update typer

mike0sv:


aguschin (approved):
Checked, this works with typer 0.4, 0.5, 0.6. Thus removing pin for typer version in `setup.py`.

PR #370, opened 2022-08-03
Remove labels and description

### LINKED ISSUE #368: Remove description and labels
Remove them to not confuse users (we have the same stuff in GTO). 

Checked: adding an extra field to a `.mlem` file doesn't cause MLEM to fail while reading the model. So this change doesn't break backward compatibility.

This will require updating Docs.

aguschin:
close #368 

I'm keeping `params` for now since I don't think it's contradictory to anything in GTO. Just some meta which users may want to save, so there should be a way to do that. Like "meta" in DVC which doesn't pretend to replace description or labels. WDYT, @mike0sv @dmpetrov

### DIFF
--- a/mlem/core/metadata.py
+++ b/mlem/core/metadata.py
@@ -50,7 +50,6 @@ def save(
     index: bool = None,
     external: Optional[bool] = None,
     params: Dict[str, str] = None,
-    update: bool = False,
 ) -> MlemObject:
     """Saves given object to a given path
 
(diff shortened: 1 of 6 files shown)


mlem/core/metadata.py
```diff
@@ -70,22 +65,18 @@ def save(
...
         params: arbitrary params for object
-        labels: labels for object
-        update: whether to keep old description/labels/params if new values were not provided
+        update: whether to keep old params if new values were not provided
 
     Returns:
         None
     """
-    if update and (description is None or params is None or labels is None):
+    if update and params is None:
```
mike0sv, line 74: i'd say let's overwrite params each time

mike0sv (approved):

mlem/core/metadata.py
aguschin: Then we don't need an `update` flag at all, right?

mike0sv: yes

PR #374, opened 2022-08-05
add support for deployment to K8s

madhur-tandon:
This also contains commits from `new POC` branch.


mlem/contrib/docker/base.py
```diff
@@ -335,7 +330,9 @@ class DockerEnv(MlemEnv[DockerContainer]):
     type: ClassVar = "docker"
     deploy_type: ClassVar = DockerContainer
     registry: DockerRegistry = DockerRegistry()
-    daemon: DockerDaemon = DockerDaemon(host="")
+    daemon: DockerDaemon = DockerDaemon(
+        host=os.getenv("DOCKER_HOST", default="")
```
mike0sv, line 334: This should go to internals of DockerDaemon. Now it overwrites `DOCKER_HOST`, make it overwrite only if actual host was passed & restore previous env value after client is created

mlem/contrib/kubernetes/base.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+class K8sDeploymentState(DeployState):
+    type: ClassVar = "kubernetes"
+    image: Optional[DockerImage]
+    pod_name: Optional[str]
+    kube_config_file_path: Optional[str] = None
+
+    def get_client(self):
+        config.load_kube_config(
```
mike0sv, line 42: seems like this is not ready?
anyway, I moved this method to `MlemDeployment` class in saegmaker PR so it has both meta and state available

mlem/contrib/kubernetes/base.py
```diff
@@ -0,0 +1,126 @@
...
+    "Running": DeployStatus.RUNNING,
+    "Succeeded": DeployStatus.STOPPED,
+    "Failed": DeployStatus.CRASHED,
+    "Unknown": DeployStatus.UNKNOWN,
+}
+
+
+class K8sDeploymentState(DeployState):
+    type: ClassVar = "kubernetes"
+    image: Optional[DockerImage]
```
mike0sv, line 37: set default values to None to be able to create empty state

mlem/contrib/kubernetes/base.py
```diff
@@ -0,0 +1,126 @@
...
+    def get_client(self):
+        config.load_kube_config(
+            config_file=self.kube_config_file_path
+            or os.getenv("KUBECONFIG", default="~/.kube/config")
+        )
+
+
+class K8sDeployment(MlemDeployment):
+    type: ClassVar = "kubernetes"
+    server: Optional[Server] = None
```
mike0sv, line 50: Better not to mix classvars with fields visually

mlem/contrib/kubernetes/base.py
```diff
@@ -0,0 +1,126 @@
...
+    def create_k8s_resources():
+        pass
+
+    def deploy(self, meta: K8sDeployment):
+        self.check_type(meta)
+        redeploy = False
+        with meta.lock_state():
+            state: K8sDeploymentState = meta.get_state()
+            if state.image is None or meta.model_changed():
+                from ..docker.helpers import build_model_image
```
mike0sv, line 69: why local import?

mlem/contrib/kubernetes/base.py
```diff
@@ -0,0 +1,126 @@
...
+
+                image_name = (
```
mike0sv, line 71: move this block to separate method?

mlem/contrib/kubernetes/base.py
```diff
@@ -0,0 +1,126 @@
...
+        self.check_type(meta)
+        state: K8sDeploymentState = meta.get_state()
+        if state.pod_name is None:
+            return DeployStatus.NOT_DEPLOYED
+
+        return DeployStatus.STOPPED  # replace it by actual value
+        # get state from kubernetes client and use POD_STATE_MAPPING to return
+
+
+class K8sYamlBuilder(DockerImageBuilder):
```
mike0sv, line 116: I guess you don't need to subclass docker image builder. In case you need only yaml youll be able to do it. In case you need both, just run `mlem build` twice

mlem/contrib/kubernetes/context.py
```diff
@@ -0,0 +1,39 @@
...
+
+from pydantic import BaseModel
+
+from mlem.utils.templates import TemplateModel
+
+logger = logging.getLogger(__name__)
+
+
+class K8sYamlBuildArgs(BaseModel):
+    name: Optional[str] = "mlem-app"
```
mike0sv, line 13: Are they really optional? Meaning can they value be `None`? 
They are not required since you provided default value, but it's not the same as `Optional`

mlem/contrib/kubernetes/context.py
```diff
@@ -0,0 +1,39 @@
...
+    image: Optional[str] = "mlem-app"
+    image_pull_policy: Optional[str] = "Always"
+    port: Optional[int] = 8080
+    service_type: Optional["str"] = "NodePort"
+
+
+class K8sYamlGenerator(K8sYamlBuildArgs, TemplateModel):
```
mike0sv, line 20: You don't need separate Args class, since you only use it once. Docker args are used multiple times if I remember correctly

mlem/contrib/kubernetes/context.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def prepare_dict(self):
+        logger.debug(
+            'Generating Resource Yaml via templates from "%s"...',
+            self.templates_dir,
+        )
+
+        logger.debug('Docker image is based on "%s".', self.image)
+
+        k8s_yaml_args = {
```
mike0sv, line 32: `self.dict()`?

mike0sv (commented):
The direction is right, now add those kubectl calls :D

mlem/contrib/kubernetes/base.py
madhur-tandon: ready now

mlem/contrib/kubernetes/base.py
madhur-tandon: done

mlem/contrib/kubernetes/base.py
madhur-tandon: hmm, this is done everywhere -- even for classes like `DockerContainer`, `DockerDirBuilder`, `DockerImageBuilder`, etc.

mlem/contrib/kubernetes/base.py
madhur-tandon: done

mlem/contrib/kubernetes/base.py
madhur-tandon: yeah, I was integrating it so that it builds docker also for us, but I guess that's not required, so fixed it and removed it. It now spits out a yaml which a person can apply manually using `kubectl`

mlem/contrib/kubernetes/base.py
madhur-tandon: hmm, I am following how it's done for Docker, so kinda keeping it if it's okay? Since Docker also follows the same structure.

mlem/contrib/kubernetes/context.py
madhur-tandon: yeah, on a 2nd thought, all are required I guess, removed `Optional`, thanks!

mlem/contrib/kubernetes/context.py
madhur-tandon: yeah, but also removing `templates_dir` while returning

mlem/contrib/docker/install_req.j2
```diff
@@ -1,4 +1,5 @@
 {% if packages %}RUN {{ package_install_cmd }} {{ packages|join(" ")  }}{% endif %}
+RUN pip install --upgrade pip
```
mike0sv, line 2: just wondering if this really neccessary

mlem/contrib/kubernetes/base.py
mike0sv: then we should move them in those classes too

mlem/contrib/kubernetes/base.py
```diff
@@ -0,0 +1,199 @@
...
+
+                sleep(0.5)
+
+                deployments_list = (
+                    client.AppsV1Api().list_namespaced_deployment(
+                        namespace=f"mlem-{meta.name}-app"
+                    )
+                )
+
+                state.deployment_name = deployments_list.items[0].metadata.name
```
mike0sv, line 125: what if it fails? or what if 0.5 seconds is not enough?
ideally, you should check if pod exists and running or throw error otherwise

mlem/contrib/docker/base.py
madhur-tandon: removed this

mlem/contrib/docker/install_req.j2
madhur-tandon: removed

mlem/contrib/kubernetes/base.py
madhur-tandon: fixed, we have a watcher now

mlem/contrib/kubernetes/base.py
mike0sv: pls fix

mlem/contrib/kubernetes/base.py
mike0sv: better to follow sagemaker since it's newer. we should refactor others the same way

mlem/contrib/kubernetes/base.py
```diff
@@ -0,0 +1,199 @@
...
+        host = service.items[0].status.load_balancer.ingress[0].ip
+        return HTTPClient(host=host, port=port)
+
+
+class K8sEnv(MlemEnv[K8sDeployment]):
+    type: ClassVar = "kubernetes"
+    deploy_type: ClassVar = K8sDeployment
+    registry: Optional[DockerRegistry] = DockerRegistry()
+
+    def create_k8s_resources(self, generator):
```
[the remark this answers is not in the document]
mike0sv: pls fix

mlem/contrib/kubernetes/base.py
mike0sv: please also check that `deployments_list.items` is not empty

mlem/contrib/kubernetes/base.py
```diff
@@ -0,0 +1,247 @@
...
+                    port=meta.port,
+                    service_type=meta.service_type,
+                )
+                generator = K8sYamlGenerator(
+                    **k8s_yaml_builder.get_resource_args().dict()
+                )
+                self.create_k8s_resources(generator)
+
+                w = watch.Watch()
+                for event in w.stream(
```
mike0sv, line 159: also a good candidate for separate method. and also make timeout configurable

mlem/contrib/kubernetes/resources.yaml.j2
```diff
@@ -0,0 +1,47 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: mlem-{{ image_name }}-app
+  labels:
+    name: mlem-{{ image_name }}-app
```
mike0sv, line 6: make it a separate field. it may default to field

mike0sv: it might even be a list?

mike0sv: and everywhere else you used this one field

mlem/contrib/kubernetes/base.py
mike0sv: also, why server is optional? can you deploy without it?

mlem/contrib/kubernetes/base.py
```diff
@@ -0,0 +1,247 @@
...
+    def find_index(self, nodes_list, node_name):
+        for i, each_node in enumerate(nodes_list):
+            if each_node.metadata.name == node_name:
+                return i
+        return -1
+
+    def _get_client(self, state: K8sDeploymentState) -> Client:
+        self.load_kube_config()
+        service = client.CoreV1Api().list_namespaced_service(
+            f"mlem-{state.deployment_name}-app"
```
mike0sv, line 69: probably should be a field and not a "dynamic" constant

mlem/contrib/kubernetes/base.py
```diff
@@ -0,0 +1,247 @@
...
+        )
+        if self.service_type == "NodePort":
+            port = service.items[0].spec.ports[0].node_port
```
mike0sv, line 72: I see 7 potential `KetError`s in this function :)

mlem/contrib/kubernetes/base.py
```diff
@@ -0,0 +1,247 @@
...
+            if each_node.metadata.name == node_name:
+                return i
+        return -1
+
+    def _get_client(self, state: K8sDeploymentState) -> Client:
+        self.load_kube_config()
+        service = client.CoreV1Api().list_namespaced_service(
+            f"mlem-{state.deployment_name}-app"
+        )
+        if self.service_type == "NodePort":
```
mike0sv, line 71: this is good opportunity to make `service_type` an `MlemABC`, esp if we plan to add others later

mlem/contrib/kubernetes/base.py
```diff
@@ -0,0 +1,247 @@
...
+            service_type=self.service_type,
+        )
+
+    def build(self, obj: MlemModel):
+        resource_args = self.get_resource_args()
+
+        generator = K8sYamlGenerator(**resource_args.dict())
+        resource_yaml = generator.generate()
+
+        generator.write("resources.yaml")
```
mike0sv, line 241: should be a field

mlem/contrib/kubernetes/base.py
```diff
@@ -0,0 +1,247 @@
...
+            image_uri=self.image.uri,
+            image_pull_policy=self.image_pull_policy,
+            port=self.port,
+            service_type=self.service_type,
+        )
+
+    def build(self, obj: MlemModel):
+        resource_args = self.get_resource_args()
+
+        generator = K8sYamlGenerator(**resource_args.dict())
```
mike0sv, line 238: doesnt make sense. both builder and yamlgenerator are subclasses of buildargs. and you create new buildargs object from current builder just to serialize it to create yamlgenerator

mike0sv: and you cant actually provide `templates_dir` for builder in current layout

mlem/contrib/kubernetes/base.py
```diff
@@ -0,0 +1,247 @@
...
+        resource_yaml = generator.generate()
+
+        generator.write("resources.yaml")
+        echo(
+            EMOJI_OK
+            + f"resources.yaml generated for {obj.basename}, apply manualy using kubectl OR use mlem deploy"
```
mike0sv, line 244: second part is probably redundant, at least we dont have anything like this in other builders

mlem/contrib/kubernetes/context.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def prepare_dict(self):
+        logger.debug(
+            'Generating Resource Yaml via templates from "%s"...',
+            self.templates_dir,
+        )
+
+        logger.debug('Docker image is based on "%s".', self.image_uri)
+
+        k8s_yaml_args = self.dict().copy()
```
mike0sv, line 32: I think you dont need copy here

tests/contrib/test_kubernetes/conftest.py
```diff
@@ -0,0 +1,152 @@
+import os
+
+import pytest
+from kubernetes import client, config
+
+from tests.conftest import long
+
+
+def _cut_empty_lines(string):
```
mike0sv, line 9: use from docker tests? maybe move to conftest

tests/contrib/test_kubernetes/test_base.py
```diff
@@ -0,0 +1,124 @@
...
+
+@pytest.fixture
+def minikube_env_variables():
+    output = subprocess.check_output(
+        ["minikube", "-p", "minikube", "docker-env"]
+    )
+    export_re = re.compile('export ([A-Z_]+)="(.*)"\\n')
+    export_pairs = export_re.findall(output.decode("UTF-8"))
+    for k, v in export_pairs:
+        os.environ[k] = v
```
mike0sv, line 33: this may pollute envs for other tests

tests/contrib/test_kubernetes/test_base.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+@pytest.fixture
+def load_kube_config():
+    config.load_kube_config(os.getenv("KUBECONFIG", default="~/.kube/config"))
```
mike0sv, line 38: isnt config for minikube constant?

mike0sv: i mean, what if someone runs tests on a machine with production kubernetes configured?

tests/contrib/test_kubernetes/test_base.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+@pytest.fixture
+def k8s_deployment(minikube_env_variables, model_meta_saved_single):
+    return K8sDeployment(
+        name="ml",
+        model=model_meta_saved_single.make_link(),
+        image_pull_policy="Never",
```
mike0sv, line 46: probably worth to make them an enum?

tests/contrib/test_kubernetes/test_base.py
```diff
@@ -0,0 +1,124 @@
...
+        service_type="LoadBalancer",
+        daemon=DockerDaemon(host=os.getenv("DOCKER_HOST", default="")),
```
mike0sv, line 48: why is that needed?

tests/contrib/test_kubernetes/test_base.py
```diff
@@ -0,0 +1,124 @@
...
+@pytest.fixture
+def docker_image(k8s_deployment):
+    tmpdir = tempfile.mkdtemp()
+    k8s_deployment.dump(os.path.join(tmpdir, "deploy"))
+    return build_model_image(
+        k8s_deployment.get_model(),
+        k8s_deployment.image_name,
+        k8s_deployment.server
+        or project_config(
+            k8s_deployment.loc.project if k8s_deployment.is_saved else None
```
mike0sv, line 61: it cant be saved since you just created it

tests/contrib/test_kubernetes/test_base.py
```diff
@@ -0,0 +1,124 @@
...
+    return K8sDeployment(
+        name="ml",
+        model=model_meta_saved_single.make_link(),
+        image_pull_policy="Never",
+        service_type="LoadBalancer",
+        daemon=DockerDaemon(host=os.getenv("DOCKER_HOST", default="")),
+    )
+
+
+@pytest.fixture
```
mike0sv, line 52: should this be session scope?

tests/contrib/test_kubernetes/test_base.py
```diff
@@ -0,0 +1,124 @@
...
+def k8s_env():
+    return K8sEnv()
+
+
+@k8s_test
+def test_deploy(
+    load_kube_config,  # pylint: disable=unused-argument
```
mike0sv, line 85: pytest.mark.usefixtures

tests/contrib/test_kubernetes/test_base.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+@k8s_test
+def test_deployed_service(
+    load_kube_config,  # pylint: disable=unused-argument
+    k8s_deployment,
+    k8s_deployment_state,
+    k8s_env,
+):
+    time.sleep(15)
```
mike0sv, line 116: ???

tests/contrib/test_kubernetes/test_base.py
```diff
@@ -0,0 +1,124 @@
...
+    k8s_deployment.update_state(k8s_deployment_state)
+    k8s_env.deploy(k8s_deployment)
+    cmd = Command("minikube tunnel")
+    cmd.run(timeout=20, shell=True)
```
mike0sv, line 120: shouldnt this be stopped after or smth like that?

tests/contrib/test_kubernetes/test_context.py
```diff
@@ -0,0 +1,33 @@
+from mlem.contrib.kubernetes.context import K8sYamlBuildArgs, K8sYamlGenerator
+from tests.contrib.test_kubernetes.conftest import _cut_empty_lines
+
+
+def test_k8s_yaml_build_args_default(k8s_default_manifest):
```
mike0sv, line 5: probably worth to move fixtures here so it's more visible (and you dont use them anywhere else)

tests/contrib/test_kubernetes/test_base.py
```diff
@@ -0,0 +1,124 @@
...
+    config.load_kube_config(os.getenv("KUBECONFIG", default="~/.kube/config"))
+
+
+@pytest.fixture
+def k8s_deployment(minikube_env_variables, model_meta_saved_single):
+    return K8sDeployment(
+        name="ml",
+        model=model_meta_saved_single.make_link(),
+        image_pull_policy="Never",
+        service_type="LoadBalancer",
```
mike0sv, line 47: need to test other value too

tests/contrib/test_kubernetes/test_base.py
madhur-tandon: yes it's stopped due to `timeout` as `20`
you are talking about `minikube tunnel` right?

madhur-tandon: so it will stop after 20 seconds, the tunneling stuff basically

mlem/contrib/kubernetes/base.py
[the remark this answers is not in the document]
madhur-tandon: done, moved it to a separate `utils.py` file

mlem/contrib/kubernetes/base.py
madhur-tandon: done

mlem/contrib/kubernetes/base.py
madhur-tandon: done

mlem/contrib/kubernetes/base.py
madhur-tandon: put it under `try/except`

mlem/contrib/kubernetes/base.py
madhur-tandon: There are only 3 types of services -- `ClusterIP`, `NodePort` and `LoadBalancer`
Out of these 3, only the last 2 are reachable from outside the Kubernetes cluster. Hence only 2 if/else.
But I have added handling of invalid values supplied.

mlem/contrib/kubernetes/context.py
madhur-tandon: that is because I am popping `templates_dir` key below and don't want to modify the original dict

tests/contrib/test_kubernetes/conftest.py
madhur-tandon: moved to `conftest` for the whole test suite

tests/contrib/test_kubernetes/test_base.py
madhur-tandon: because our `daemon` needs to be aware of `minikube's docker host` which was set as with `os.environ` in a fixture. That's why `DOCKER_HOST` value is used inside the test, but not in the original field / flow.

mlem/contrib/kubernetes/base.py
madhur-tandon: separated visually is done, I guess one can deploy it because of 
```
meta.server or project_config(meta.loc.project if meta.is_saved else None).server
```

where this server (from the 2nd part) comes from
```
@property
    def server(self):
        from mlem.runtime.server import Server

        if not self.SERVER:
            return parse_obj_as(Server, {"type": "fastapi"})
        return parse_obj_as(Server, self.SERVER)
```

so it will work even if `meta.server` is None

mlem/contrib/kubernetes/base.py
madhur-tandon: done

tests/contrib/test_kubernetes/test_base.py
madhur-tandon: hmm, what do you suggest then? I have restored it's state after `yield` for now. Hopefully, that should suffice.

tests/contrib/test_kubernetes/test_base.py
madhur-tandon: hmm, by default, all kubeconfig files are stored in the same path i.e. `~/.kube/config`
doesn't matter if it's minikube or EKS
so if your kubeconfig points to EKS and you run the tests, then yes, tests will be run on EKS

tests/contrib/test_kubernetes/test_base.py
madhur-tandon: done

tests/contrib/test_kubernetes/test_context.py
madhur-tandon: done

tests/contrib/test_kubernetes/test_base.py
madhur-tandon: we can, but it will have a cascading effect i.e. other fixtures will have to also be made of `session` scope -- basically the one it depends on such as `k8s_deployment` and then `minikube_env_variables` which is ok, but the problem comes from `model_meta_saved_single` which we don't want to make it as session scope I guess.

so, not doing it

tests/contrib/test_kubernetes/test_base.py
madhur-tandon: removed, passing only `None` now

mlem/contrib/kubernetes/resources.yaml.j2
madhur-tandon: this was due yes, made a new one called `{{ namespace }}` now

mlem/contrib/kubernetes/base.py
madhur-tandon: this is also fixed as a side-effect of creating `{{ namespace }}`

mlem/contrib/kubernetes/base.py
madhur-tandon: removed `apply manualy using kubectl OR use mlem deploy`

mlem/contrib/kubernetes/base.py
mike0sv: resource_yaml

tests/contrib/test_kubernetes/test_base.py
madhur-tandon: this was because when the 2 tests ran one after the other, it took some time to delete the resources (namespace, deployments, etc.) created by the first test. Otherwise it would complain that things already exist, since 2nd test tries to create stuff again.

But, I have now fixed this and am using a waiting mechanism using a watcher that waits until the resources are cleared. So this is fixed now.

mlem/contrib/kubernetes/base.py
mike0sv: target like for docker_dir

mlem/contrib/kubernetes/context.py
mike0sv: `.dict()` already creates new dict because `self` is a model, not a dict

tests/contrib/test_kubernetes/test_base.py
mike0sv: yeah and that's wrong, tests should always run on minikube

mlem/contrib/kubernetes/base.py
```diff
@@ -0,0 +1,255 @@
...
+
+
+class K8sDeployment(MlemDeployment, K8sYamlBuildArgs):
+    """MlemDeployment implementation for Kubernetes deployments"""
+
+    type: ClassVar = "kubernetes"
+
+    server: Optional[Server] = None
+    """type of Server to use, with options such as FastAPI, RabbitMQ etc."""
+    state_type: ClassVar = K8sDeploymentState
```
mike0sv, line 57: this is also a classvar, move it up

madhur-tandon: missed it, will fix

mlem/contrib/kubernetes/base.py
```diff
@@ -0,0 +1,255 @@
...
+            elif self.service_type == "LoadBalancer":
+                port = service.items[0].spec.ports[0].port
+                ingress = service.items[0].status.load_balancer.ingress[0]
+                host = ingress.hostname or ingress.ip
+            else:
+                raise ValueError(
+                    f"service_type supplied is {self.service_type}, valid values are [ClusterIP, NodePort, LoadBalancer] only"
+                )
+        except (IndexError, ValueError, TypeError) as e:
+            print(
```
mike0sv, line 106: better to raise from e here

mlem/contrib/kubernetes/base.py
```diff
@@ -0,0 +1,255 @@
...
+                address_dict = node_list[node_index].status.addresses
+                for each_address in address_dict:
+                    if each_address.type == "ExternalDNS":
+                        host = each_address.address
+            elif self.service_type == "LoadBalancer":
+                port = service.items[0].spec.ports[0].port
+                ingress = service.items[0].status.load_balancer.ingress[0]
+                host = ingress.hostname or ingress.ip
+            else:
+                raise ValueError(
```
mike0sv, line 102: this will raise if it's clusterip
create an mlem abc with abstract method get_host_and_port

mlem/contrib/kubernetes/base.py
```diff
@@ -0,0 +1,255 @@
...
+
+    deploy_type: ClassVar = K8sDeployment
+    """type of deployment being used for the Kubernetes environment"""
+    registry: Optional[DockerRegistry] = None
+    """docker registry"""
+
+    def get_registry(self, meta: K8sDeployment):
+        if self.registry is None and meta.registry is None:
+            echo(EMOJI_FAIL + "Error: No registry set to be used by Docker")
+            raise ValueError(
```
mike0sv, line 130: should be mlem error. and you wont need an echo before this

mlem/contrib/kubernetes/base.py
madhur-tandon: done

mlem/contrib/kubernetes/base.py
madhur-tandon: done

mlem/contrib/kubernetes/base.py
madhur-tandon: done

mlem/contrib/kubernetes/context.py
madhur-tandon: done

tests/contrib/test_kubernetes/test_base.py
madhur-tandon: hmm, the mechanism of how EKS works and minikube works is different. Even if I determine host and port correctly in minikube for `nodeport`, one still cannot access it because it's inside a VM.

so we need to use a separate command that spits out a URL
and since we cannot get the relevant client, the only way to test it would be using something like `requests` (i.e. equivalent to `curl`)

one of the reasons this is complicated is because `minikube` runs only one node i.e. the control plane and not the worker nodes.

so given all the complexity, I would suggest to skip this.

mlem/contrib/kubernetes/base.py
madhur-tandon: done the mlem abc thing, thinking of raising an error to handle ClusterIP case

madhur-tandon: raised an error, done

mlem/contrib/kubernetes/base.py
madhur-tandon: done

mlem/contrib/kubernetes/base.py
madhur-tandon: done

mlem/contrib/kubernetes/base.py
```diff
@@ -0,0 +1,281 @@
...
+    def get_host_and_port(
+        self, service, namespace="mlem"  # pylint: disable=unused-argument
+    ) -> Tuple[Optional[str], Optional[int]]:
+        """Returns host and port for the service in Kubernetes"""
+        return None, None
+
+
+class NodePortService(ServiceType):
+
+    type: ClassVar = "k8s_nodeport_service"
```
mike0sv, line 64: you can just use "nodeport", "loadbalancer" etc since abs_type already has k8s service type keywords

mlem/contrib/kubernetes/base.py
```diff
@@ -0,0 +1,281 @@
...
+
+    def get_host_and_port(self, service, namespace="mlem"):
+        host, port = None, None
```
mike0sv, line 67: remove?

mlem/contrib/kubernetes/base.py
```diff
@@ -0,0 +1,281 @@
...
+            client.CoreV1Api()
+            .list_namespaced_pod(namespace)
+            .items[0]
+            .spec.node_name
+        )
+        node_list = client.CoreV1Api().list_node().items
+        node_index = find_index(node_list, node_name)
+        address_dict = node_list[node_index].status.addresses
+        for each_address in address_dict:
+            if each_address.type == "ExternalDNS":
```
mike0sv, line 79: if there is no externaldns, will the host be valid? if not, we should raise error

mlem/contrib/kubernetes/base.py
madhur-tandon: done

mlem/contrib/kubernetes/base.py
madhur-tandon: I will change it to `ExternalIP` and then raise and error if not found.

mlem/contrib/kubernetes/base.py
```diff
@@ -0,0 +1,281 @@
...
+            config_file=self.kube_config_file_path
+            or os.getenv("KUBECONFIG", default="~/.kube/config")
+        )
+
+    def _get_client(self, state: K8sDeploymentState) -> Client:
+        host, port = None, None
+        self.load_kube_config()
+        service = client.CoreV1Api().list_namespaced_service(self.namespace)
+        try:
+            if self.service_type == ServiceTypeEnum.node_port:
```
mike0sv, line 136: you should make `service_type` a `ServiceType` field and just call `self.service_type.get_host_and_port()`

mlem/contrib/kubernetes/base.py
```diff
@@ -0,0 +1,281 @@
...
+            elif self.service_type == ServiceTypeEnum.cluster_ip:
+                raise TypeError(
+                    "Cannot expose service of type ClusterIP outside the Kubernetes Cluster"
+                )
+            else:
+                raise ValueError(
+                    f"service_type supplied is {self.service_type}, valid values are {[e.value for e in ServiceTypeEnum]} only"
+                )
+        except (IndexError, ValueError, TypeError) as e:
+            print("Couldn't determine host and port from the service deployed")
```
mike0sv, line 151: Do not use prints, everything should be in exception message. If you need something else, raise different exception from this

mlem/contrib/kubernetes/base.py
```diff
@@ -0,0 +1,281 @@
...
+                host, port = LoadBalancerService().get_host_and_port(service)
+            elif self.service_type == ServiceTypeEnum.cluster_ip:
+                raise TypeError(
+                    "Cannot expose service of type ClusterIP outside the Kubernetes Cluster"
+                )
+            else:
+                raise ValueError(
+                    f"service_type supplied is {self.service_type}, valid values are {[e.value for e in ServiceTypeEnum]} only"
+                )
+        except (IndexError, ValueError, TypeError) as e:
```
mike0sv, line 150: Those checks should be inside corresponding `get_host_and_port` implementations, not common. Those index errors can mean different things for different implementations and even different things for one implementation

mlem/contrib/kubernetes/base.py
```diff
@@ -0,0 +1,281 @@
...
+    """docker registry"""
+    daemon: Optional[DockerDaemon] = DockerDaemon(host="")
+    """docker daemon"""
+    kube_config_file_path: Optional[str] = None
+    """path for kube config file of the cluster"""
+    templates_dir: List[str] = []
+    """list of dirs where templates reside"""
+
+    def load_kube_config(self):
+        config.load_kube_config(
```
mike0sv, line 126: does this set config globally? we should not pollute users configuration if there is any. is there a way to not load this globally?

mlem/contrib/kubernetes/base.py
```diff
@@ -0,0 +1,281 @@
...
+                    daemon=meta.daemon,
+                    server=self.get_server(meta),
+                )
+                meta.update_model_hash(state=state)
+                redeploy = True
+
+            if state.deployment_name is None or redeploy:
+                generator = K8sYamlGenerator(
+                    namespace=meta.namespace,
+                    image_name=state.image.name,  # type: ignore
```
mike0sv, line 212: why # type: ignore?

madhur-tandon: because `state.image` is `Optional[DockerImage]` which complains

```
Item "None" of "Optional[DockerImage]" has no attribute "name"
```

mlem/contrib/kubernetes/base.py
```diff
@@ -0,0 +1,281 @@
...
+                create_k8s_resources(generator)
+
+                if pod_is_running(namespace=meta.namespace):
+                    deployments_list = (
+                        client.AppsV1Api().list_namespaced_deployment(
+                            namespace=meta.namespace
+                        )
+                    )
+
+                    if len(deployments_list.items):
```
mike0sv, line 228: should we raise error otherwise?

mlem/contrib/kubernetes/base.py
madhur-tandon: As in? The config already is present in a user's system, this command just makes the K8s API aware of it. So we are not really polluting anything. We are just using what is already present. And with each new run of the python process / our CLI tool, config has to be loaded again and again. Different invocations won't remember that we already loaded it before.

mlem/contrib/kubernetes/base.py
madhur-tandon: Do you recommend raising `e` again?

mlem/contrib/kubernetes/base.py
madhur-tandon: `DeploymentError` should be fine i guess? But we also raise it below. We can raise it here also but with a different message.

madhur-tandon: did that

mlem/contrib/kubernetes/base.py
madhur-tandon: done, used a property instead to calculate the right type

mlem/contrib/kubernetes/base.py
madhur-tandon: done

mlem/contrib/kubernetes/base.py
madhur-tandon: Index errors and attribute errors are common really. They don't mean different things. Type Error and Value error is what I added because I was explicitly raising `TypeError` for ClusterIP service and ValueError when the service type was something different than what is allowed in `Enum`

mlem/contrib/kubernetes/base.py
mike0sv: ...

mlem/contrib/kubernetes/base.py
```diff
@@ -0,0 +1,300 @@
...
+        node_index = find_index(node_list, node_name)
+        if node_index == -1:
+            raise DeploymentError(
+                f"Couldn't find the node where pods in namespace {namespace} exists"
+            )
+        address_dict = node_list[node_index].status.addresses
+        external_ip_found = False
+        for each_address in address_dict:
+            if each_address.type == "ExternalIP":
+                host = each_address.address
```
mike0sv, line 86: you can just return here and raise instead of current return

mlem/contrib/kubernetes/base.py
```diff
@@ -0,0 +1,300 @@
...
+        return None, None
+
+
+class NodePortService(ServiceType):
+
+    type: ClassVar = "nodeport"
+
+    def get_host_and_port(self, service, namespace="mlem"):
+        host, port = None, None
+        port = service.items[0].spec.ports[0].node_port
```
mike0sv, line 69: Another 2 IndexErrors

mlem/contrib/kubernetes/base.py
```diff
@@ -0,0 +1,300 @@
...
+        node_name = (
+            client.CoreV1Api()
+            .list_namespaced_pod(namespace)
+            .items[0]
```
mike0sv, line 73: IndexError

mlem/contrib/kubernetes/base.py
```diff
@@ -0,0 +1,300 @@
...
+            )
+        return host, port
+
+
+class LoadBalancerService(ServiceType):
+
+    type: ClassVar = "loadbalancer"
+
+    def get_host_and_port(self, service, namespace="mlem"):
+        host, port = None, None
```
mike0sv, line 100: ...

mlem/contrib/kubernetes/base.py
```diff
@@ -0,0 +1,300 @@
...
+        port = service.items[0].spec.ports[0].port
```
mike0sv, line 101: 2 IndexErrors

mlem/contrib/kubernetes/base.py
```diff
@@ -0,0 +1,300 @@
...
+        ingress = service.items[0].status.load_balancer.ingress[0]
```
mike0sv, line 102: another 2 IndexErrors

mlem/contrib/kubernetes/base.py
```diff
@@ -0,0 +1,300 @@
...
+        host = ingress.hostname or ingress.ip
+        return host, port
+
+
+class ClusterIPService(ServiceType):
+
+    type: ClassVar = "clusterip"
+
+    def get_host_and_port(self, service, namespace="mlem"):
+        raise TypeError(
```
mike0sv, line 112: Why type error?

mike0sv: Also all exceptions not subclassed from MlemError are treated as unexpected. So this should be MlemError

mlem/contrib/kubernetes/base.py
madhur-tandon: yes these are handled commonly when this method is used inside the `try` block of `_get_client`

mlem/contrib/kubernetes/base.py
madhur-tandon: okay, thanks, will do

mlem/contrib/kubernetes/base.py
madhur-tandon: the index errors are handled commonly in the `try` block of `_get_client`

mlem/contrib/kubernetes/base.py
madhur-tandon: the index errors are handled commonly in the `try` block of `_get_client`

mlem/contrib/kubernetes/base.py
mike0sv: If user calls mlem api from his code while simultaneously working with `kubernetes` python package with configuration loaded from different source, you will pollute his configuration

mlem/contrib/kubernetes/base.py
mike0sv: That means you should check if it not `None`, not just ignore the warning

mlem/contrib/kubernetes/base.py
madhur-tandon: the index errors are handled commonly in the `try` block of `_get_client`

mlem/contrib/kubernetes/base.py
mike0sv: Now it's to far apart to be readable :( can you invert conditions?
like
```
if not ...:
   raise ...
... continue with what was under if block
```

tests/contrib/test_kubernetes/test_base.py
mike0sv: but that means you building a docker for every test, which is very long. I think there were other model meta fixtures that are session-scoped. Or you can create your own

tests/contrib/test_kubernetes/test_base.py
mike0sv: so the test will always run for 20 seconds?

tests/contrib/test_kubernetes/test_base.py
madhur-tandon: I am creating my own then

tests/contrib/test_kubernetes/test_base.py
madhur-tandon: I think test should stop immediately once it passes. The 20 seconds stuff is for minikube tunnel to stop, we don't want to be waiting for prediction forever, so it stops the tunneling in the background after 20 seconds which is enough for test to pass and stop.

mlem/contrib/kubernetes/base.py
madhur-tandon: fixed

mlem/contrib/kubernetes/base.py
madhur-tandon: done

mlem/contrib/kubernetes/base.py
madhur-tandon: done

mlem/contrib/kubernetes/base.py
madhur-tandon: done

mlem/contrib/kubernetes/base.py
madhur-tandon: done

mlem/contrib/kubernetes/base.py
madhur-tandon: done

mlem/contrib/kubernetes/base.py
madhur-tandon: done

mlem/contrib/kubernetes/base.py
madhur-tandon: done

mlem/contrib/kubernetes/base.py
madhur-tandon: done

mlem/contrib/kubernetes/base.py
mike0sv: create an issue for that, we'll do it separately

mlem/contrib/kubernetes/context.py
```diff
@@ -0,0 +1,59 @@
...
+logger = logging.getLogger(__name__)
+
+
+class ImagePullPolicy(str, Enum):
+    always = "Always"
+    never = "Never"
+    if_not_present = "IfNotPresent"
+
+
+class ServiceTypeEnum(str, Enum):
```
mike0sv, line 19: Do you still need this?

madhur-tandon: hmm, not really, but it gives a signal that there are limited types available
and of course will error out if we supply an invalid value.
So I am slightly inclined to keep it

mike0sv (approved):

PR #389, opened 2022-08-30
Add link to the weekly board to README

aguschin:


aguschin:
@omesser, could you please check if this looks good?

omesser:
@aguschin  - lgtm!
Only suggestion - maybe link to the current week view (because you're calling it "weekly board" in the link)

PR #391, opened 2022-09-01
create MlemSource to choose how mlem is added to docker

### LINKED ISSUE #345: Run `apt-get update` before installing unix dependencies
In docker build, some unix dependencies will fail to install unless apt is updated first. We need to
1. if unix dependencies are present, run update before installing them
2. use best practices for it like doing updating and installation in one step, clear apt cache in the same step to make layer size smaller
3. add option to docker build context to be able to alter updating and clearing commands to support other base images

mike0sv:
close https://github.com/iterative/mlem/issues/345

### DIFF
--- a/mlem/cli/dev.py
+++ b/mlem/cli/dev.py
@@ -4,7 +4,7 @@ from mlem.cli.main import app, mlem_command, mlem_group
 from mlem.ui import echo
 from mlem.utils.entrypoints import (
     MLEM_ENTRY_POINT,
-    find_implementations,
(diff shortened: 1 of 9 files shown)


aguschin:
close https://github.com/iterative/mlem/issues/345 if I'm correct. Right @mike0sv ?

mike0sv:
Lol yes
I forgot that we have it already as an issue :)

aguschin (approved):
LGTM!

PR #392, opened 2022-09-01
add reqs for non-standard sklearn io

mike0sv:


aguschin:
Which issue does this resolve?

mike0sv:
It was from discord message. Not an issue, I just thought that someday it may fail and fixed

aguschin:
Ok, got it. There is a test failing still, in case you missed it.

PR #395, opened 2022-09-02
Get rid of mlem dir

### LINKED ISSUE #381: MLEM index and `.mlem/` folder
Feedback from @daavoo

> Why store everything in `.mlem`? How can I choose the output folder?
I discovered `external` in project structure page, perhaps worth explicit mention?

This brings us back to the discussion of having an index in MLEM. I know @shcheklein and @mike0sv had conversations about this. So, have you decided something on this? From what I recall:

1. In GS we should delete executing `$ mlem init` to untie MLEM usage from MLEM project (we have a ticket for it in mlem.ai repo)
2. To support `mlem ls` without `.mlem/` we need to implement repo parsing (just like DVC I suppose)

mike0sv:
closes #381


mlem/api/commands.py
```diff
@@ -202,14 +189,12 @@ def clone(
...
-        index=index,
-        external=external,
     )
 
 
 def init(path: str = ".") -> None:
-    """Creates .mlem directory in `path`"""
-    path = posixpath.join(path, MLEM_DIR)
+    """Creates mlem config in `path`"""
+    path = posixpath.join(path, MLEM_CONFIG_FILE_NAME)
```
aguschin, line 197: I know the file is called `.mlem.yaml`. But for GTO it's `.gto`. Should we rename it to `.gto.yaml` in GTO, or should we rename `.mlem.yaml` in `.mlem`? We shouldn't confuse users with different approaches. Did we had same files in other tools maybe? cc @shcheklein

aguschin (approved):
LGTM! close #381

mlem/api/commands.py
aguschin: @mike0sv if everything else is good, let's merge this so we won't get a PR queue again? And then discuss it separately.

omesser: my 2 cents - if the file is indeed a `yaml`, `gto.yaml` > `[something.]gto`
extensions are for formats :)
as long as `gto` doesn't invent a new file format/structure but respects `yaml`, that's the [least surprising](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) thing to do

aguschin: Ok, then it's decided I guess. Let's keep it here, and I'll update it in GTO as well some time later.

PR #398, opened 2022-09-09
Fix docstrings

mike0sv:
Address docsting feedback from #363


mlem/cli/apply.py
```diff
@@ -187,13 +190,17 @@ def _apply_remote(
 
 
 option_output = Option(
-    None, "-o", "--output", help="Where to store the outputs."
+    None,
+    "-o",
+    "--output",
+    help="Where to save inference results",
```
jorgeorpinel, line 196: Are model outputs always inferences? Idk. Maybe something more specific like "model's output files"? Up to you

mlem/cli/main.py
```diff
@@ -481,20 +488,32 @@ def inner(*iargs, **ikwargs):
...
 )
+option_model = Option(
+    ...,
+    "-m",
+    "--model",
+    help="Path to MLEM model object",
+    metavar=PATH_METAVAR,
+)
+option_data = Option(
+    ..., "-d", "--data", help="Path to MLEM data object", metavar=PATH_METAVAR
```
jorgeorpinel, line 515: Clearer 👍🏼 though maybe "model/data MLEM object" would be a better phrasing (keep "MLEM object" together). No strong opinion

mlem/cli/declare.py
```diff
@@ -53,15 +53,15 @@ def create_declare_mlem_object_subcommand(
...
+        section="MLEM Objects",
         parent=parent,
         dynamic_metavar="__kwargs__",
         dynamic_options_generator=abc_fields_parameters(subtype, parent_cls),
         hidden=subtype.startswith("_"),
         lazy_help=lazy_class_docstring(type_name, subtype),
     )
     def subtype_command(
-        path: str = Argument(..., help="Where to save object"),
+        path: str = Argument(..., help="Where to save the object"),
```
jorgeorpinel, line 64: Specify "(.mlem file)" ?

mlem/cli/serve.py
```diff
@@ -17,15 +18,15 @@
...
 )
 from mlem.core.metadata import load_meta
 from mlem.core.objects import MlemModel
 from mlem.runtime.server import Server
 
 serve = Typer(
     name="serve",
-    help="""Deploy the model locally using a server implementation and expose its methods as
-endpoints.
+    help="""Create an API from model methods using a server implementation.
```
jorgeorpinel, line 29: "REST server" ?

jorgeorpinel (approved):

mlem/cli/main.py
mike0sv: @aguschin wdyt?

mlem/cli/serve.py
mike0sv: It depends on implementation. Some of them not even http-based

mlem/cli/main.py
aguschin: IMO `MLEM model object` is simpler than `model MLEM object`.

shcheklein: I think it should be "Path to MLEM model" (ppl don't care about "object", it doesn't give additional information).

shcheklein: Is it .mlem path or path to an actual directory / file that MLEM creates? MLEM model object ideally should be just "Path to model"

jorgeorpinel: I think it's the path to the .mlem file which is why I suggested somewhere else to add `(.mlem file)` just to be specific.

aguschin: No, it's a path to an actual model. E.g. if you have `model.pkl` and `model.pkl.mlem` nearby, the argument should be `model.pkl` (although it should work if one supplies `model.pkl.mlem` as well).

The thing is, if there is no `model.pkl.mlem`, MLEM won't understand what it is. So it should be MLEM object (== `model.pkl.mlem` should exist) for this to work, not just a binary saved with pickle/joblib or csv file saved with pandas.

mike0sv: What @aguschin said is correct, I just see it the other way around - you should specify path to `.mlem` file, but omitting `.mlem` suffix also works. But if somehow you have `aaa.mlem` that uses `bbb.pkl` file, you can't specify `bbb.pkl` - you can use `aaa.mlem` or `aaa`. But generally `bbb.pkl` == `aaa`, unless you tamper with mlem files manually.
As for what @shcheklein said: I added `MLEM Object` to imply that it only works with models saved by MLEM. We can omit it, but I think this will make it less clear. We can also do `Path to MLEM model` probably

aguschin: `Path to MLEM model` and `Path to MLEM dataset` sounds good to me.

jorgeorpinel: I see this is merged but just FYI this is similar to `dvc push` (and other commands) which `targets` can be either tracked files/dirs or .dvc files. See e.g. https://dvc.org/doc/command-reference/push for ideas on good phrasing around this.

PR #410, opened 2022-09-16
Fix/stderr

### LINKED ISSUE #404: Write errors to `stderr` instead of `stdout`
Looks like error messages are written to stdout:

```shell
$ mlem ls https://
❌ .mlem folder wasn't found when searching through the path. Search has started from here: path=https://,
fs=<fsspec.implementations.http.HTTPFileSystem object at 0x105fdb7f0>, rev=None
export X=`mlem ls https://`
$ echo $X
❌ .mlem folder wasn't found when searching through the path. Search has started from here: path=https://,
fs=<fsspec.implementations.http.HTTPFileSystem object at 0x10fff4910>, rev=None
```

itstargetconfirmed:
Decided Approach: 

I noticed that exceptions thrown and eventually caught in `mlem/cli/main.py` utilize the `cli_echo` context manager and the `echo` function to print exceptions to the console. 

The `cli_echo` context manager sets the `echo` function to print to `stdout` through a `Console` instance provided by the Rich package. By default, `Console` will write to `stdout`, hence the reason why exceptions are printed to `stdout`. 

I did some research and discovered that the Rich package supports consoles that can print to `stderr` by passing `stderr=True` when creating the `Console`. With this knowledge, I created a `Console` instance called `error_console` with `stderr=True`. If an error needs to be printed to `stderr`, we can utilize the `error_console` and if we need to print to `stdout`, the existing `console` which prints to `stdout` can be used. 

I then introduced a new context manager called `stderr_echo` that will switch the `echo` function’s underlying print function to `error_console.print` when an error needs to be printed. If printing to `stdout` is required, then we will utilize the exisitng `cli_echo` context manager that will switch the `echo` function’s underlying print function to `console.print`.

With this approach, we use the existing Rich package for printing to both `stdout` and `stderr` thus providing consistency. Additionally, it allows us to keep the signature of the `echo` function the same and let Rich manage printing to different streams under the hood. 

Alternative Approaches: 

I thought about other approaches to implement printing to `stderr`, for example, using `sys.stderr.write` directly. The problem with this approach is that the `echo` function currently excepts multiple parameters including a `Text` instance (along with other configurations) specific to the Rich package, however `sys.stderr.write` requires only a string parameter. 

If we were to use this approach, we’d need to write code to only accept a string parameter (but not a `Text` instance) if we’re printing an error. It is possible through function overloading, however, it’s definitely a cleaner implementation to handling this through the Rich package. 

Related Tasks: 

If this approach is accepted and merged, I will update the contribution guide to explain the usage of the `cli_echo` and `stderr_echo` context managers along with the `echo` function. 

Testing: 

With the ability to print to both `stdout` and `stderr`, it is important that we are able to develop tests that are capable of reading and differentiating outputs from both these streams. 

In the current cli test module, the `CliRunner` instance writes both `stdout` and `stderr` streams to `result.output`. If a test needs to differentiate where the output came from, it is unable to do so. 

I’ve performed a couple updates to the cli test module. 

1.	I’ve updated the `CliRunner` instance by using `mix_stderr=False`, which outputs the `stderr` stream to `result.stderr` and outputs the `stdout` stream to `result.stdout`.  
2.	I’ve updated all tests in the cli test module to use `result.stderr` and/or `result.stdout` for their `assert` statements.
3.	I’ve update cli tests so that error messages for `assert` statements will now print the `stdout` stream, `stderr` stream and exceptions by using `(result.stdout, result.stderr, result.exception)`. 
4.	I’ve implemented tests to ensure that `MlemError` and `Exception` are printed to stderr. 

close https://github.com/iterative/mlem/issues/404

### DIFF
--- a/mlem/cli/main.py
+++ b/mlem/cli/main.py
@@ -23,7 +23,15 @@ from mlem.core.errors import MlemError
 from mlem.core.metadata import load_meta
 from mlem.core.objects import MlemObject
 from mlem.telemetry import telemetry
-from mlem.ui import EMOJI_FAIL, EMOJI_MLEM, bold, cli_echo, color, echo
+from mlem.ui import (
+    EMOJI_FAIL,
+    EMOJI_MLEM,
+    bold,
+    cli_echo,
+    color,
+    echo,
+    stderr_echo,
+)
 
 
 class MlemFormatter(HelpFormatter):
@@ -289,7 +297,7 @@ def mlem_command(
                 error = f"{e.__class__.__module__}.{e.__class__.__name__}"
                 if ctx.obj["traceback"]:
                     raise
-                with cli_echo():
+                with stderr_echo():
                     echo(EMOJI_FAIL + color(str(e), col=typer.colors.RED))
                 raise typer.Exit(1)
             except ValidationError as e:
@@ -297,14 +305,14 @@ def mlem_command(
                 if ctx.obj["traceback"]:
                     raise
                 msgs = "\n".join(_format_validation_error(e))
-                with cli_echo():
+                with stderr_echo():
                     echo(EMOJI_FAIL + color("Error:\n", "red") + msgs)
                 raise typer.Exit(1)
             except Exception as e:  # pylint: disable=broad-except
                 error = f"{e.__class__.__module__}.{e.__class__.__name__}"
                 if ctx.obj["traceback"]:
                     raise
-                with cli_echo():
+                with stderr_echo():
                     echo(
                         EMOJI_FAIL
                         + color(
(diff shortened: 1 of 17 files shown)


aguschin (approved):
Thanks @itstargetconfirmed! Looks good to me! Merged `main` branch from upstream into yours to make the test pass (@mike0sv was fixing some stuff in CI already after you made a fork).

Also please note that CI don't start unless anyone from @iterative/mlem approves it, so once you push something, please don't hesitate to ping us to trigger the CI :)

itstargetconfirmed:
Hi @aguschin @mike0sv, I saw codecov threw a warning because the code coverage was 88.88%, however, the target amount is 89.39%. It suggested that stderr_echo should be covered in the unit tests. I've updated the PR to include a test for stderr_echo.

aguschin:
Ok, I see the test of DVC extensions fails. Fixing that test. Since most of CI jobs were passing except for that test and few jobs that didn't have time to complete, I'm merging this.

PR #414, opened 2022-09-21
Fix/verbose mode

itstargetconfirmed:
Decided Approach: 

Typer implements a callback that allows users to provide parameters for the main cli application, including printing the application version, printing a traceback and printing logs through verbose mode. 

In reviewing the implementation for printing a traceback, it uses the `traceback` property on the context object to track if the user enabled traceback printing. The `traceback` property on the context object is enabled by the either explicitly setting the `--trackback` option or setting the `DEBUG` property of `LOCAL_CONFIG` to true. I’ve included the relevant source code below: 

`ctx.obj = {"traceback": traceback or LOCAL_CONFIG.DEBUG}`

We can easily print a traceback through the `--verbose` option by checking if the verbose option was provided in the cli and changing the state of the `traceback` property accordingly (shown below).

`ctx.obj = {"traceback": traceback or verbose or LOCAL_CONFIG.DEBUG}`

User Experience / Messaging Updates

I’ve also implemented the suggestion by @mike0sv to include a message to use the `--tb` or `--traceback` option in the case of unexpected errors. I’ve included the error message below: 

`Use the --tb or --traceback option to include the traceback in the output`

I've also added a similar message and workflow for `MlemError` and `ValidationError` since they raise exceptions as well. 

I also added tests for `--traceback` and `--verbose` options to ensure that exception data is sent to stderr when exceptions are thrown. These tests are located in `tests/cli/test_stderr.py` since we're checking the contents of stderr. 

I also needed to make a couple modifications to a few existing tests since they checked for raised exceptions, instead, I've modified them to look for `raise [Exception Type]` in stderr. 

Things I'm Unsure About

After I updated  `ctx.obj = {"traceback": traceback or verbose or LOCAL_CONFIG.DEBUG}` to check the value of `verbose` the `tests/contrib/test_onnx.py::test_model_type__dump_load` fails. I can't put by finger on why this is occurring.

### DIFF
--- a/mlem/cli/main.py
+++ b/mlem/cli/main.py
@@ -1,5 +1,6 @@
 import contextlib
 import logging
+import traceback
 import typing as t
 from collections import defaultdict
 from enum import Enum, EnumMeta
@@ -31,6 +32,10 @@ from mlem.ui import (
     color,
     echo,
     stderr_echo,
+)
+
+TRACEBACK_SUGGESTION_MESSAGE = (
+    "Use the --tb or --traceback option to include the traceback in the output"
 )
 
 
@@ -246,7 +251,7 @@ def mlem_callback(
         logger = logging.getLogger("mlem")
         logger.handlers[0].setLevel(logging.DEBUG)
         logger.setLevel(logging.DEBUG)
-    ctx.obj = {"traceback": traceback or LOCAL_CONFIG.DEBUG}
+    ctx.obj = {"traceback": traceback or verbose or LOCAL_CONFIG.DEBUG}
 
 
 def _extract_examples(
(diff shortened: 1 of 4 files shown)


mlem/cli/main.py
```diff
@@ -484,3 +492,10 @@ def config_arg(
             obj = build_mlem_object(model, subtype, conf, file_conf)
 
     return obj
+
+
+def handle_traceback():
+    with stderr_echo():
+        traceback_output = traceback.format_exc()
+        echo(traceback_output)
+    raise typer.Exit(1)
```
mike0sv, line 501: Not sure this is needed, python exception traceback already should go to stderr (`python -c "raise Exception()" > /dev/null` will print stderr). And we are loosing exception information (you needed to fix tests because of it)

aguschin:
Hi @itstargetconfirmed! Would you like to finish this one?

mlem/cli/main.py
aguschin: @mike0sv, should we just finish this? Looks like @itstargetconfirmed almost did it.

PR #416, opened 2022-09-23
use cloudpickle for saving torch models

### LINKED ISSUE #415: Failed to load simple torch model saved with MLEM
Hi I'm trying to get to know mlem.
I'm facing weird error.
Nothing goes wrong if I run the following script.
```python
import torch
from mlem.api import save, load


class MyNet(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.l1 = torch.nn.Linear(10, 5)
        self.l2 = torch.nn.Linear(5, 1)

    def forward(self, x):
        return self.l2(self.l1(x))

net = MyNet()
# Save and load mlem
saved = save(net, 'mlem-test')
loaded = load('mlem-test')
```
But If I try to load model again in another terminal. I get attribute error.
```python
from mlem.api import load

loaded = load('mlem-test')
```
Error log
```shell
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Input In [2], in <cell line: 1>()
----> 1 loaded = load('mlem-test')

File /usr/local/lib/python3.9/dist-packages/mlem/core/metadata.py:100, in load(path, project, rev, batch_size, follow_links)
     81 def load(
     82     path: str,
     83     project: Optional[str] = None,
   (...)
     86     follow_links: bool = True,
     87 ) -> Any:
     88     """Load python object saved by MLEM
     89 
     90     Args:
   (...)
     98         Any: Python object saved by MLEM
     99     """
--> 100     meta = load_meta(
    101         path,
    102         project=project,
    103         rev=rev,
    104         follow_links=follow_links,
    105         load_value=batch_size is None,
    106     )
    107     if isinstance(meta, MlemData) and batch_size:
    108         return meta.read_batch(batch_size)

File /usr/local/lib/python3.9/dist-packages/mlem/core/metadata.py:180, in load_meta(path, project, rev, follow_links, load_value, fs, force_type)
    175 meta = cls.read(
    176     location=find_meta_location(location),
    177     follow_links=follow_links,
    178 )
    179 if load_value:
--> 180     meta.load_value()
    181 if not isinstance(meta, cls):
    182     raise WrongMetaType(meta, force_type)

File /usr/local/lib/python3.9/dist-packages/mlem/core/objects.py:627, in MlemModel.load_value(self)
    625 def load_value(self):
    626     with self.requirements.import_custom():
--> 627         self.model_type.load(self.relative_artifacts)

File /usr/local/lib/python3.9/dist-packages/mlem/core/model.py:249, in ModelType.load(self, artifacts)
    248 def load(self, artifacts: Artifacts):
--> 249     self.model = self.io.load(artifacts)

File /usr/local/lib/python3.9/dist-packages/mlem/contrib/torch.py:154, in TorchModelIO.load(self, artifacts)
    152 load = torch.jit.load if self.is_jit else torch.load
    153 with artifacts[self.art_name].open() as f:
--> 154     return load(f)

File /usr/local/lib/python3.9/dist-packages/torch/serialization.py:712, in load(f, map_location, pickle_module, **pickle_load_args)
    710             opened_file.seek(orig_position)
    711             return torch.jit.load(opened_file)
--> 712         return _load(opened_zipfile, map_location, pickle_module, **pickle_load_args)
    713 return _legacy_load(opened_file, map_location, pickle_module, **pickle_load_args)

File /usr/local/lib/python3.9/dist-packages/torch/serialization.py:1049, in _load(zip_file, map_location, pickle_module, pickle_file, **pickle_load_args)
   1047 unpickler = UnpicklerWrapper(data_file, **pickle_load_args)
   1048 unpickler.persistent_load = persistent_load
-> 1049 result = unpickler.load()
   1051 torch._utils._validate_loaded_sparse_tensors()
   1053 return result

File /usr/local/lib/python3.9/dist-packages/torch/serialization.py:1042, in _load.<locals>.UnpicklerWrapper.find_class(self, mod_name, name)
   1040         pass
   1041 mod_name = load_module_mapping.get(mod_name, mod_name)
-> 1042 return super().find_class(mod_name, name)

AttributeError: Can't get attribute 'MyNet' on <module '__main__'>

```
What am I doing wrong?
Any help would be great.

Environment and versions are
Ubuntu 20.04
mlem==0.2.8
torch==1.12.1+cu116

madhur-tandon:
Fixes https://github.com/iterative/mlem/issues/415

I will check and update the PR if `cloudpickle` can be used in other places also.

### DIFF
--- a/mlem/contrib/torch.py
+++ b/mlem/contrib/torch.py
@@ -152,10 +152,9 @@ class TorchModelIO(ModelIO):
         if len(artifacts) != 1:
             raise ValueError("Invalid artifacts: should have only one file")
 
+        load = torch.jit.load if self.is_jit else torch.load
         with artifacts[self.art_name].open() as f:
-            if self.is_jit:
-                return torch.jit.load(f)
-            return torch.load(f, pickle_module=cloudpickle)
+            return load(f)
 
 
 class TorchModel(ModelType, ModelHook, IsInstanceHookMixin):
--- a/mlem/contrib/torch.py
+++ b/mlem/contrib/torch.py
@@ -1,5 +1,6 @@
 from typing import Any, ClassVar, Iterator, List, Optional, Tuple
 
+import cloudpickle
 import torch
 from pydantic import conlist, create_model
 
@@ -140,18 +141,21 @@ class TorchModelIO(ModelIO):
 
     def dump(self, storage: Storage, path, model) -> Artifacts:
         self.is_jit = isinstance(model, torch.jit.ScriptModule)
-        save = torch.jit.save if self.is_jit else torch.save
         with storage.open(path) as (f, art):
-            save(model, f)
+            if self.is_jit:
+                torch.jit.save(model, f)
+            else:
+                torch.save(model, f, pickle_module=cloudpickle)
             return {self.art_name: art}
 
     def load(self, artifacts: Artifacts):
         if len(artifacts) != 1:
             raise ValueError("Invalid artifacts: should have only one file")
 
-        load = torch.jit.load if self.is_jit else torch.load
         with artifacts[self.art_name].open() as f:
-            return load(f)
+            if self.is_jit:
+                return torch.jit.load(f)
+            return torch.load(f, pickle_module=cloudpickle)
 
 
 class TorchModel(ModelType, ModelHook, IsInstanceHookMixin):
--- a/tests/contrib/test_torch.py
+++ b/tests/contrib/test_torch.py
@@ -18,6 +18,7 @@ from mlem.core.data_type import DataAnalyzer, DataType
 from mlem.core.errors import DeserializationError, SerializationError
 from mlem.core.model import ModelAnalyzer
 from mlem.core.objects import MlemModel
+from mlem.utils.path import make_posix
 from tests.conftest import data_write_read_check
 
 
@@ -178,14 +179,14 @@ def test_torch_import(tmp_path, net, torchsave):
 
 def test_torch_import_in_separate_shell():
     tmp_path = mkdtemp()
-    path = os.path.join(tmp_path, "model")
+    path = make_posix(os.path.join(tmp_path, "model"))
     m = torch.nn.Linear(5, 1)
     torch.save(m, path)
     x = subprocess.run(
         [
             "python",
             "-c",
-            f"from mlem.api import import_object; from mlem.contrib.torch import TorchModelImport; import_object('{str(path)}', type_=TorchModelImport.type)",
+            f"from mlem.api import import_object; from mlem.contrib.torch import TorchModelImport; import_object('{path}', type_=TorchModelImport.type)",
         ],
         shell=True,
         check=True,
--- a/tests/contrib/test_torch.py
+++ b/tests/contrib/test_torch.py
@@ -183,7 +183,7 @@ def test_torch_import_in_separate_shell(tmp_path):
         [
             "python",
             "-c",
-            f"from mlem.api import load; loaded = load('{path}')",
+            f""""from mlem.api import load; loaded = load('{path}')" """,
         ],
         check=True,
     )
--- a/tests/contrib/test_torch.py
+++ b/tests/contrib/test_torch.py
@@ -191,6 +191,8 @@ def test_torch_import_in_separate_shell():
         check=True,
     )
     assert x.returncode == 0
+    os.remove(path)
+    os.rmdir(tmp_path)
 
 
 # Copyright 2019 Zyfra
--- a/tests/contrib/test_torch.py
+++ b/tests/contrib/test_torch.py
@@ -1,6 +1,5 @@
 import os
 import subprocess
-from tempfile import mkdtemp
 
 import pytest
 import torch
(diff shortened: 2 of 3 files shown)


mlem/contrib/torch.py
```diff
@@ -140,9 +141,11 @@ class TorchModelIO(ModelIO):
 
     def dump(self, storage: Storage, path, model) -> Artifacts:
         self.is_jit = isinstance(model, torch.jit.ScriptModule)
-        save = torch.jit.save if self.is_jit else torch.save
         with storage.open(path) as (f, art):
-            save(model, f)
+            if self.is_jit:
+                torch.jit.save(model, f)
+            else:
+                torch.save(model, f, pickle_module=cloudpickle)
```
mike0sv, line 148: can you also try `dill` here? since we already use it

tests/contrib/test_torch.py
```diff
@@ -174,6 +177,25 @@ def test_torch_import(tmp_path, net, torchsave):
     assert isinstance(meta.model_type, TorchModel)
 
 
+def test_torch_import_in_separate_shell():
+    tmp_path = mkdtemp()
```
mike0sv, line 181: use fixture instead

tests/contrib/test_torch.py
```diff
@@ -174,6 +177,25 @@ def test_torch_import(tmp_path, net, torchsave):
...
+    path = make_posix(os.path.join(tmp_path, "model"))
+    m = torch.nn.Linear(5, 1)
+    torch.save(m, path)
+    x = subprocess.run(
+        [
+            "python",
+            "-c",
+            f"from mlem.api import import_object; from mlem.contrib.torch import TorchModelImport; import_object('{path}', type_=TorchModelImport.type)",
```
mike0sv, line 189: probably easier to use mlem cli?

tests/contrib/test_torch.py
```diff
@@ -174,6 +177,25 @@ def test_torch_import(tmp_path, net, torchsave):
     assert isinstance(meta.model_type, TorchModel)
 
 
+def test_torch_import_in_separate_shell():
+    tmp_path = mkdtemp()
+    path = make_posix(os.path.join(tmp_path, "model"))
+    m = torch.nn.Linear(5, 1)
```
mike0sv, line 183: does this break without cloudpickle?

tests/contrib/test_torch.py
madhur-tandon: We want a temp dir that persists between different shells / terminals.
If I use the existing fixture, that doesn't persist.

So we need to use `mkdtemp ` which is not automatically deleted.
I can make a new fixture that uses `mkdtemp`, does that work?

tests/contrib/test_torch.py
madhur-tandon: what breaks? declaration of `m = torch.nn.Linear(5, 1)`? Nope.
only loading from a separate terminal breaks as described in the issue here: https://github.com/iterative/mlem/issues/415

mlem/contrib/torch.py
madhur-tandon: you mean instead of `cloudpickle`?

mike0sv: Yes

tests/contrib/test_torch.py
mike0sv: Yes, I mean the entire test. Original problem was with custom net class which you do not use in this test, so I'm asking if you testing the same case

tests/contrib/test_torch.py
mike0sv: WDYM `persists between different shells`?

mike0sv: it's just a path. and you generate and unlink it in the same test instance

madhur-tandon: By persist, I mean that, when I run the subprocess call, that new shell / terminal should be able to access the tempdir.

By default, dirs created by the `tempfile` library delete themselves automatically unless one uses `mkdtemp`

tests/contrib/test_torch.py
madhur-tandon: Aah yes, you are right, this will pass in both cases, I will change to a custom net here.

mlem/contrib/torch.py
madhur-tandon: `dill` didn't work, using `cloudpickle` only

tests/contrib/test_torch.py
madhur-tandon: done

tests/contrib/test_torch.py
madhur-tandon: done

tests/contrib/test_torch.py
madhur-tandon: as in? I have made it simpler btw, but still didn't use `mlem cli` stuff

aguschin (approved):
LGTM!

tests/contrib/test_torch.py
aguschin: I assume Mike meant something like `mlem pprint {path}`, but don't see big problem with this solution)

PR #417, opened 2022-09-23
Refactoring of deployments and env

mike0sv:
This is last bits on refactoring deployments. It includes:

- move deploying logic from envs to deployments
- remove model field from deployment, now you should provide it every time you run deployment
- changes to `mlem deployment run` similar to those in #363 
- add used declaration to state to detect changes (eg change in heroku app name). For now we should just forbid changes (ie just create new deployment and remove old one). In future we can allow changing some of the fields (scaling factor, for example)
- add used model info to state. For now it will be only path and project (if used explicitly). In future we should also add rev for git-tracked models (commit sha, not branch/tag) to link deployment state with repo "state". Also a good idea will be to have an option to forbid deploying not tracked models (ie changed after last commit) or at least warn about it. This will allow users to enforce this linking and have exactly the same version of model in commit and in deployment

Some examples of how deployments cli can be used:
```
# declare everything separately
mlem declare env heroku --api_key lol prod.mlem
mlem declare deployment heroku --env prod.mlem --app_name myapp service.mlem
mlem deployment run --load service.mlem --model mdoel
# ----
# declare deployment with embedded env
mlem declare deployment heroku --env.api_key lol --app_name myapp service.mlem
mlem deployment run --load service.mlem --model mdoel
# ----
# declare env only
mlem declare env heroku --api_key lol prod.mlem
mlem deployment run heroku service.mlem --model model --app_name myapp --env prod.mlem
# ---
# provide all parameters to `run
mlem deployment run heroku service.mlem --model model --app_name myapp --env.api_key lol
# error on args mismatch
```


madhur-tandon:
> * move deploying logic from envs to deployments

Had a quick look, this is basically reversing the roles of envs and deployments.
But I think tests have to be changed right now? to accommodate this reversal of roles.

> * remove model field from deployment, now you should provide it every time you run deployment

I guess this only applies to `mlem deployment run`, what about other sub-commands such as `mlem deployment apply`? They shouldn't be affected right? or `--model` has to be used with each sub-command as well?

Also, just to confirm, if I understand correctly, I guess the commands will change in such a manner right?:
BEFORE:
```
mlem deployment run testsk --model sk-model --env kubernetes
```

AFTER:
```
mlem deployment run kubernetes testsk.mlem --model sk-model
```

mike0sv:
> But I think tests have to be changed right now? 

They are.

Every command except `run` works with already existing deployment so you do not need to provide model option.

Now you can either create new deployment with `mlem deployment run <type> <path> [options]`
or run existing with `mlem deployment run --load <path>`

mlem/cli/deployment.py
```diff
@@ -64,21 +87,75 @@ def deploy_run(
...
+        index: bool = option_index,
+        file_conf: List[str] = option_file_conf("deployment"),
+        **__kwargs__,
+    ):
+        from mlem.api.commands import deploy
+
+        try:
+            meta = load_meta(path, project=project, force_type=MlemDeployment)
+            raise DeploymentError(
+                f"Deployment meta already exists at {meta.loc}. Please use mlem deployment run --load <path> ..."
```
aguschin, line 134: Should we quote the commands in errors? like 
```suggestion
                f"Deployment meta already exists at {meta.loc}. Please use `mlem deployment run --load <path> ...`"
```

mlem/contrib/docker/base.py
```diff
@@ -323,57 +348,103 @@ def ensure_image_name(self):
...
-    """Default registry to push images to"""
-    daemon: DockerDaemon = DockerDaemon(host="")
-    """Docker daemon parameters"""
+            echo(EMOJI_OK + f"Container {state.container_name} is up")
 
-    def delete_image(self, image: DockerImage, force: bool = False, **kwargs):
-        with self.daemon.client() as client:
-            return image.delete(client, force=force, **kwargs)
+    def remove(self):
+        # self.check_type(meta)
```
aguschin, line 386: ?
```suggestion
```

mlem/contrib/docker/base.py
```diff
@@ -323,57 +348,103 @@ def ensure_image_name(self):
...
+                    container = client.containers.get(state.container_id)
+                    container.stop()
+                    container.remove()
+                except docker.errors.NotFound:
+                    pass
+            state.container_id = None
+            self.update_state(state)
+
+    def get_status(self, raise_on_error=True) -> DeployStatus:
+        # self.check_type(meta)
```
aguschin, line 405: ?
```suggestion
```

mlem/contrib/heroku/meta.py
```diff
@@ -53,11 +62,12 @@ def ensured_app(self) -> HerokuAppMeta:
         return self.app
 
 
-class HerokuDeployment(MlemDeployment):
+class HerokuDeployment(MlemDeployment[HerokuState, HerokuEnv]):
```
aguschin, line 65: confused how to read this type hint. Why it could be `MlemDeployment[Y, Z]`? Could you please explain?

aguschin (approved):
LGTM! 🚀

mlem/contrib/heroku/meta.py
mike0sv: This is for mypy and IDE help mostly. `MlemDeployment` is `Generic` and methods like `get_env` will return not arbitrary `MlemEnv`, but whatever you provide as second generic parameter (`HerokuEnv` in this case).

PR #421, opened 2022-10-02
add separate reader/writer for lgb labels

### LINKED ISSUE #224: Save label for LightGBM datasets as separately
Labels are a part of `lgb.Dataset`, but not a part of `inner` DatasetType. So we need to add separate field `lables: DatasetType` to DatasetType and corresponding field to Reader/Writer and make it create separate artifact for labels

madhur-tandon:
Closes https://github.com/iterative/mlem/issues/224

### DIFF
--- a/mlem/contrib/lightgbm.py
+++ b/mlem/contrib/lightgbm.py
@@ -45,27 +45,23 @@ class LightGBMDataType(
     type: ClassVar[str] = "lightgbm"
     valid_types: ClassVar = (lgb.Dataset,)
     inner: DataType
-    labels: Optional[DataType]
+    labels: DataType
 
     def serialize(self, instance: Any) -> dict:
         self.check_type(instance, lgb.Dataset, SerializationError)
         res = {
-            LIGHTGBM_DATA: self.inner.get_serializer().serialize(instance.data)
-        }
-        if self.labels is not None:
-            res[LIGHTGBM_LABEL] = self.labels.get_serializer().serialize(
+            LIGHTGBM_DATA: self.inner.get_serializer().serialize(
+                instance.data
+            ),
+            LIGHTGBM_LABEL: self.labels.get_serializer().serialize(
                 instance.label
-            )
+            ),
+        }
         return res
 
     def deserialize(self, obj: dict) -> Any:
         data = self.inner.get_serializer().deserialize(obj[LIGHTGBM_DATA])
-        if self.labels is not None:
-            label = self.labels.get_serializer().deserialize(
-                obj[LIGHTGBM_LABEL]
-            )
-        else:
-            label = None
+        label = self.labels.get_serializer().deserialize(obj[LIGHTGBM_LABEL])
         try:
             return lgb.Dataset(data, label=label, free_raw_data=False)
         except ValueError as e:
@@ -87,6 +83,7 @@ class LightGBMDataType(
 
     @classmethod
     def process(cls, obj: Any, **kwargs) -> DataType:
+        obj = obj.construct()
         return LightGBMDataType(
             inner=DataAnalyzer.analyze(obj.data),
             labels=DataAnalyzer.analyze(obj.label),
@@ -107,26 +104,21 @@ class LightGBMDataWriter(DataWriter):
                 f"expected data to be of LightGBMDataType, got {type(data)} instead"
             )
 
-        res = {}
-        lightgbm_construct = data.data.construct()
-        raw_data = lightgbm_construct.get_data()
-        raw_label = lightgbm_construct.get_label().tolist()
+        lightgbm_raw = data.data
 
         inner_reader, inner_art = data.inner.get_writer().write(
-            data.inner.copy().bind(raw_data),
+            data.inner.copy().bind(lightgbm_raw.data),
             storage,
             posixpath.join(path, LIGHTGBM_DATA),
         )
-        res[LIGHTGBM_DATA] = inner_art
 
-        labels_reader = None
-        if data.labels is not None:
-            labels_reader, labels_art = data.labels.get_writer().write(
-                data.labels.copy().bind(raw_label),
-                storage,
-                posixpath.join(path, LIGHTGBM_LABEL),
-            )
-            res[LIGHTGBM_LABEL] = labels_art
+        labels_reader, labels_art = data.labels.get_writer().write(
+            data.labels.copy().bind(lightgbm_raw.label),
+            storage,
+            posixpath.join(path, LIGHTGBM_LABEL),
+        )
+
+        res = {LIGHTGBM_DATA: inner_art, LIGHTGBM_LABEL: labels_art}
 
         return (
             LightGBMDataReader(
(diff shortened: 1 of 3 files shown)


madhur-tandon:
Other tests that fail are unrelated, this should be ready.

mlem/contrib/lightgbm.py
```diff
@@ -42,15 +44,28 @@ class LightGBMDataType(
...
     def serialize(self, instance: Any) -> dict:
         self.check_type(instance, lgb.Dataset, SerializationError)
-        return self.inner.get_serializer().serialize(instance.data)
+        res = {"inner": self.inner.get_serializer().serialize(instance.data)}
+        if instance.label:
+            res["labels"] = self.labels.get_serializer().serialize(
+                instance.label
+            )
+        else:
+            res["labels"] = PrimitiveType(ptype="NoneType").serialize(
```
mike0sv, line 57: seems redundant?

mlem/contrib/lightgbm.py
```diff
@@ -70,7 +85,12 @@ def get_writer(
 
     @classmethod
     def process(cls, obj: Any, **kwargs) -> DataType:
-        return LightGBMDataType(inner=DataAnalyzer.analyze(obj.data))
+        return LightGBMDataType(
+            inner=DataAnalyzer.analyze(obj.data),
+            labels=DataAnalyzer.analyze(obj.label)
```
mike0sv, line 90: Make `labels` datatype optional?

mlem/contrib/lightgbm.py
```diff
@@ -86,33 +106,51 @@ def write(
...
         raw_data = lightgbm_construct.get_data()
-        underlying_labels = lightgbm_construct.get_label().tolist()
-        inner_reader, art = data.inner.get_writer().write(
-            data.inner.copy().bind(raw_data), storage, path
+        raw_label = lightgbm_construct.get_label().tolist()
+
+        inner_reader, inner_art = data.inner.get_writer().write(
+            data.inner.copy().bind(raw_data),
+            storage,
+            posixpath.join(path, "inner"),
```
mike0sv, line 117: `inner` -> `data`?

mike0sv: and make those strings constants

mlem/contrib/lightgbm.py
```diff
@@ -86,33 +106,51 @@ def write(
...
+        )
+        labels_reader, labels_art = data.labels.get_writer().write(
+            data.labels.copy().bind(raw_label),
+            storage,
+            posixpath.join(path, "labels"),
         )
+
+        res["inner"] = inner_art
+        res["labels"] = labels_art
```
mike0sv, line 126: can replace with inplace dict creation

mlem/contrib/lightgbm.py
```diff
@@ -86,33 +106,51 @@ def write(
...
-        inner_reader, art = data.inner.get_writer().write(
-            data.inner.copy().bind(raw_data), storage, path
+        raw_label = lightgbm_construct.get_label().tolist()
+
+        inner_reader, inner_art = data.inner.get_writer().write(
+            data.inner.copy().bind(raw_data),
+            storage,
+            posixpath.join(path, "inner"),
+        )
+        labels_reader, labels_art = data.labels.get_writer().write(
```
mike0sv, line 119: Can we do old logic if no labels present?
Without labels: `save(data, "dataset")` produces
```
dataset.mlem
dataset
```
With labels:
```
dataset.mlem
dataset/data
dataset/labels
```

mlem/core/data_type.py
```diff
@@ -230,6 +230,9 @@ def read(self, artifacts: Artifacts) -> DataType:
                 data = None
             elif res == "False":
                 data = False
+            elif self.data_type.ptype == "int":
+                # this handles cases such as int('0.0')
+                data = self.data_type.to_type(float(res))
```
mike0sv, line 235: add tests for this case?

mlem/contrib/lightgbm.py
madhur-tandon: ok

mlem/contrib/lightgbm.py
madhur-tandon: done

mlem/contrib/lightgbm.py
madhur-tandon: checking on `self.labels` now

mlem/contrib/lightgbm.py
madhur-tandon: done

mlem/contrib/lightgbm.py
madhur-tandon: I think I made it work, unfortunately, lgb tests do not work for me locally, will check on CI

madhur-tandon: Hmm, change of implementation, using `.construct()` now
this ensures that labels are never None and always exist

madhur-tandon: We will always have labels now, resolving

mlem/contrib/lightgbm.py
madhur-tandon: removed, always have labels now

mlem/contrib/lightgbm.py
madhur-tandon: always have it, so not optional now

mlem/contrib/lightgbm.py
madhur-tandon: resolving

mlem/contrib/lightgbm.py
madhur-tandon: resolving

mlem/core/data_type.py
madhur-tandon: I have removed this, this shouldn't be an issue
The fact that we were dealing with labels being automatically converted into arrays even when passed as None initially was the need for this condition, but I have removed it now

mlem/contrib/lightgbm.py
mike0sv: We should not modify the object passed to `process`. And we should always have exactly the same object when we `load` something we `save`d before. Is this the case if you are using `construct`?

madhur-tandon: When you instantiate a `lgb.Dataset`, you cannot call `get_data` and `get_label` on it since it is lazy initialized i.e. it hasn't been constructed. We were using `get_data` before this PR also i.e. we were using `.construct` in the `writer`.

Further, when you pass `labels` as None, the `.construct()` operation sets them to some values i.e. it is never None after it has been constructed.

The good thing about `.construct` is that when it is called on an already `constructed` object -- it makes no difference i.e. it can be called infinitely more times.

And methods like `lgb.train` also construct it under the hood.

So do you want to work with the raw object OR the constructed object.
If the RAW object, then I will have to change the PR so that `.construct` is never called anywhere aka the existing implementation of this PR was also kinda wrong then.

If the Constructed object, then this PR is kinda ready as it is.

madhur-tandon: I have removed the usage of `construct` now

mike0sv: Ok, now it looks a lot better - basically labels are just a second dataset.
Now again, if you save an `lgb.Dataset` without labels, can we  do it with a single file like I proposed earlier?
Basically if labels are not `None` this should be the same logic. If they are, you'll have `labels` as `None` (in reader/writer as well) and just one artifact at `<dataset>` instead of `<dataset>/data`

madhur-tandon: I have accommodated that `labels` should not be saved when it is `None`.
But then, the key with it is saved as, is still `<dataset>/data`
Do you really wanna have `<dataset>` instead of `<dataset>/data`? I just ask because it will lead to many if/else branches I guess.

madhur-tandon: nvm, I made it `<dataset>` instead of `<dataset>/data` 😄

mike0sv (approved):

madhur-tandon:
Thanks @mike0sv, merging!

Issue #330, opened 2022-07-01
Allow easier CI runs for PRs from forks


aguschin:
Currently, we need to manually approve each commit in PR from fork to be run in CI. This adds a lag into getting valuable feedback for external contributors.

Suggestion: split tests and CI into two parts: one that don't require secrets and the other that does. Then run the first each time, but still require manual approve to run the second part.

PR #433, opened 2022-10-07
run heroku tests on detection from title OR schedule

madhur-tandon:


aguschin (commented):
Good work! Trying to understand how this should work. Will it require commit hash when one tries to run the tests manually? Also, can I test it out in this branch before merging to `main`?

UPD: I just did a commit assuming it will ask me for input (commit hash), but looks like this works differently.

aguschin:
@madhur-tandon could use your help here to understand this ^ :)

.github/workflows/check-heroku-tests.yml
```diff
@@ -0,0 +1,83 @@
+name: Heroku Tests
+
+on:
+  workflow_dispatch:
+    inputs:
+      commit_hash:
+        description: 'Hash of git commit to run heroku tests on'
+        required: true
+  schedule:
+  - cron: '0 7 * * 1'  # M H d m w (Mondays at 7:00)
```
mike0sv, line 10: We have the same schedule in current wf. Can we not duplicate it and just detect if WF was triggered by schedule? Also we talked about having some specific string in commit message to trigger heroku tests even in not-scheduled runs

.github/workflows/check-heroku-tests.yml
```diff
@@ -0,0 +1,83 @@
...
+        GITHUB_USERNAME: ${{ secrets.GH_USERNAME }}
+        GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
+        AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+        AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+        HEROKU_API_KEY: ${{ secrets.HEROKU_API_KEY }}
+        HEROKU_TEAM: iterative-sandbox
+        GITHUB_MATRIX_OS: ${{ matrix.os }}
+        GITHUB_MATRIX_PYTHON: ${{ matrix.python }}
+        BITBUCKET_USERNAME: ${{ secrets.BITBUCKET_USERNAME }}
+        BITBUCKET_PASSWORD: ${{ secrets.BITBUCKET_PASSWORD }}
```
aguschin, line 78: Since we're creating a separate workflow here, maybe we can also separate the tests that don't require all these secrets? Otherwise when an external contributor pushes the PR, we have to confirm each time so it can be run. This delays the execution. If that does look simple enough, could we do it here? Otherwise we should do it separate. This is the issue https://github.com/iterative/mlem/issues/330

madhur-tandon: Using the same workflow now

.github/workflows/check-heroku-tests.yml
madhur-tandon: Doing a check on title of the PR now

madhur-tandon:
> @madhur-tandon could use your help here to understand this ^ :)

Earlier, I was using something known as a `workflow dispatch` -- i.e. you need to trigger an action manually.
But now, I have switched to detection of the keyword `heroku` in title of the PR

madhur-tandon:
The PRs https://github.com/iterative/mlem/pull/457 and https://github.com/iterative/mlem/pull/458 confirm that this works 😄

aguschin:
Looking good! Thanks @madhur-tandon!

I remember we had a discussion about doing the same for k8s/sagemaker. Don't remember the resolution. Should we do this for them also?

aguschin (approved):

.github/workflows/check-test-release.yml
```diff
@@ -114,6 +114,17 @@ jobs:
...
         BITBUCKET_USERNAME: ${{ secrets.BITBUCKET_USERNAME }}
         BITBUCKET_PASSWORD: ${{ secrets.BITBUCKET_PASSWORD }}
+    - name: Run Heroku tests
+      if: contains(github.event.pull_request.title, 'heroku')
+      timeout-minutes: 40
+      run: pytest tests/contrib/test_heroku.py
+      env:
+        GITHUB_USERNAME: ${{ secrets.GH_USERNAME }}
+        GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
+        HEROKU_API_KEY: ${{ secrets.HEROKU_API_KEY }}
```
aguschin, line 124: Ah, you didn't delete `HEROKU_API_KEY` and `HEROKU_TEAM` for the step running `pytest -k 'not heroku'`. I guess you should.

madhur-tandon: done, but don't merge it yet, will retest those 2 dummy PRs I opened

.github/workflows/check-test-release.yml
```diff
@@ -102,18 +102,27 @@ jobs:
...
         AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
         AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-        HEROKU_API_KEY: ${{ secrets.HEROKU_API_KEY }}
-        HEROKU_TEAM: iterative-sandbox
         GITHUB_MATRIX_OS: ${{ matrix.os }}
         GITHUB_MATRIX_PYTHON: ${{ matrix.python }}
         BITBUCKET_USERNAME: ${{ secrets.BITBUCKET_USERNAME }}
         BITBUCKET_PASSWORD: ${{ secrets.BITBUCKET_PASSWORD }}
+    - name: Run Heroku tests
+      if: ${{ github.event_name == 'schedule' || contains(github.event.pull_request.title, 'heroku') }}
```
aguschin, line 116: Will this work if it's `Heroku`, but not `heroku`?
```suggestion
      if: ${{ github.event_name == 'schedule' || contains(github.event.pull_request.title, 'heroku') }}
```

madhur-tandon: `This function is not case sensitive.` -- from https://docs.github.com/en/actions/learn-github-actions/expressions#contains

.github/workflows/check-heroku-tests.yml
madhur-tandon: schedule or title check -- should work on both now

madhur-tandon:
> I remember we had a discussion about doing the same for k8s/sagemaker. Don't remember the resolution. Should we do this for them also?

K8s test are self contained, won't be necessary.
@mike0sv can comment on Sagemaker maybe

aguschin:
> K8s test are self contained, won't be necessary.

how much time do they take? Speeding up CI would be a good thing also.

madhur-tandon:
> how much time do they take? Speeding up CI would be a good thing also.

The point of this PR was not to speed up CI, it was to control flaky behaviour of heroku so we don't get unnecessary failures.

Issue #390, opened 2022-08-31
Disable pretty-printing and emojis by default


aguschin:
Got this feedback from @dmpetrov (typer pretty-printing and emojis) and @shcheklein (pretty-printing only).

Some considerations:
1. All Iterative CLI tools should behave the same. If we want to use this, let's use in all tools.
2. Sometimes pretty-printing fails (@shcheklein posted some example in Slack I believe)
3. Having emojis in output may make it harder to parse stdout/stderr.
4. This is not a usual pattern for CLI

IMHO: although I like emojis and pretty-printing, I don't see this as a big question that needs spending time on, so we can just disable those. Would be nice to get some thoughts from users: maybe they like it, then we can consider using this in all Iterative products.

aguschin:
reported by @madhur-tandon [here](https://github.com/iterative/mlem/pull/434#discussion_r1001403190):

> I have replaced it with `" "`, but `pip install $(mlem build requirements -m model)` doesn't work because of the first line with the emoji i.e. `⏳️ Loading model from sk-model.mlem`.
Is there a way to ignore all this `echo` related stuff?

E.g. it looks like
```shell
$ mlem build model requirements
⏳️ Loading model from model.mlem
scikit-learn==1.0.2
pandas==1.4.2
numpy==1.22.3
```

This actually relates to the [discussion we had in GTO](https://github.com/iterative/gto/issues/107). There we decided that the output should be human-readable first, machine-readable second. That's why we implemented `--plain` and `--json` options to provide output in machine-readable format, when needed. Here I see few options:
1. by default, we make the output less verbose in all MLEM commands
2. we make the output less verbose in some commands (for now in the example above only) (don't like this option cause of non-consistency and ad-hoc fixing - we don't solve the problem holistically)
3. we add `--plain` (or another similar option) to skip the verbose output

WDYT guys? @madhur-tandon @shcheklein @mike0sv @skshetry

jorgeorpinel:
> by default, we make the output less verbose in all MLEM commands

➕ 

Currently it makes it look kind of complicated, almost like Terraform or some complex orchestration tool.

aguschin:
Regarding verbosity.
I suggest to split this in three levels:
1. Verbose. Require `-v/--verbose`. Be as verbose, as possible, with a  Basically, this is what we have now.
2. Default.
3. Quiet. Require `-q/--quiet`. Nothing printed except for the command result, if there is one.

[...]

PR #434, opened 2022-10-07
add requirements builder

### LINKED ISSUE #209: Setting up venv from `model.mlem` meta
As @AlexIoannides suggested, there should be a way to install requirements from model metadata. 

Something like
```
$ mlem pack my-model requirements
catboost==1.0.0
numpy==2.0.0

$ mlem pack my-model requiments -c target requirements.txt
< output written to requirements.txt >
```
More advanced option would be to create venv right from mlem command: `$ mlem pack my-model venv -c target my-venv`, or create envs with Pipenv or Poetry.

madhur-tandon:
Closes https://github.com/iterative/mlem/issues/209

- Using pipenv is not recommended: See https://github.com/pypa/pipenv/issues/4058
- Using subprocess to use `pip` as indicated by https://pip.pypa.io/en/latest/user_guide/#using-pip-from-your-program
- Using `conda` through subprocess and not the Conda Python API since that introduces a new dependency (which is not really installable by pip + conda list through the API doesn't include the `pypi` packages)


aguschin:
Re [the issue you posted about pipenv](https://github.com/pypa/pipenv/issues/4058) - the issue was closed in 2020, so it's a bit outdated. If you check, they are releasing [quite frequently now](https://github.com/pypa/pipenv/releases). So don't think the project is dead, unless I'm missing something.

madhur-tandon:
> Re [the issue you posted about pipenv](https://github.com/pypa/pipenv/issues/4058) - the issue was closed in 2020, so it's a bit outdated. If you check, they are releasing [quite frequently now](https://github.com/pypa/pipenv/releases). So don't think the project is dead, unless I'm missing something.

You can check numerous posts on the internet:
- https://chriswarrick.com/blog/2018/07/17/pipenv-promises-a-lot-delivers-very-little/
- https://pipenv.pypa.io/en/latest/diagnose/
- https://towardsdatascience.com/common-errors-and-how-to-solve-them-why-wont-it-lock-8f5e57111f23
- https://webdevelop.pro/blog/rejection-dependency-manager-pipenv/

aguschin:
Ok, what I read in your links:
1. some people think `pipenv` is not useful - well, there are people who like and use pipenv as well, you can check it in `used by: 39k` [here](https://github.com/pypa/pipenv)
2. `pipenv` have some problems and there are workarounds and troubleshooting guide

I would like to note that I'm not insisting on implementing something in particular (like `pipenv`, `poetry` or `conda`). The only thing that is must be implemented is obviously `virtualenv`. The others are optional to my mind. But I don't see reasons to reject `pipenv` implementation completely (let's say in some future if we don't want to implement it here).

I would also consider implementing `virtualenv` only for now to get some feedback before we go into advanced tools like `conda`, `pipenv`, `poetry` or something else.

madhur-tandon:
> I personally would consider implementing `virtualenv` only for now to get some feedback before we go into advanced tools like `conda`, `pipenv`, `poetry` or something else.

I have done `virtualenv` and `conda` as of now in this PR

mlem/contrib/requirements.py
```diff
@@ -0,0 +1,37 @@
...
+
+    type: ClassVar = "requirements"
+
+    target: Optional[str] = None
+    """Target path for the requirements.txt file"""
+
+    def write_requirements_file(self, requirements: Requirements):
+        requirement_string = "\n".join(requirements.to_pip())
+        if self.target is None:
+            print(requirement_string)
```
mike0sv, line 25: Add a test for this case

mlem/contrib/venv.py
```diff
@@ -0,0 +1,173 @@
...
+    def build(self, obj: MlemModel):
+        if self.current_env:
+            if (
+                os.getenv("VIRTUAL_ENV") is None
+                or sys.prefix == sys.base_prefix
+            ):
+                raise MlemError("No virtual environment detected.")
+            echo(EMOJI_PACK + f"Detected the virtual env {sys.prefix}")
+            env_dir = sys.prefix
+            env_exe = os.path.join(env_dir, "bin", "python")
```
mike0sv, line 82: windows case?

mlem/contrib/venv.py
```diff
@@ -0,0 +1,173 @@
...
+
+        return context
+
+    def get_installed_packages(self, env_dir):
+        if platform.system() == "Windows":
+            env_exe = os.path.join(env_dir, "Scripts", "python.exe")
+        else:
+            env_exe = os.path.join(env_dir, "bin", "python")
+
+        return subprocess.check_output([env_exe, "-m", "pip", "freeze"])
```
mike0sv, line 71: handle errors?

mlem/contrib/venv.py
```diff
@@ -0,0 +1,173 @@
...
+            env_dir = context.env_dir
+            os.environ["VIRTUAL_ENV"] = context.env_dir
+            env_exe = context.env_exe
+        echo(EMOJI_PACK + "Installing the required packages...")
+        # Based on recommendation given in https://pip.pypa.io/en/latest/user_guide/#using-pip-from-your-program
+        install_cmd = [env_exe, "-m", "pip", "install"]
+        if self.no_cache:
+            install_cmd.append("--no-cache-dir")
+        install_cmd.extend(obj.requirements.to_pip())
+        subprocess.run(install_cmd, check=True)
```
mike0sv, line 95: handle errors?

mlem/contrib/venv.py
```diff
@@ -0,0 +1,173 @@
...
+            + f"virtual environment `{self.target}` is ready, activate with `source {self.target}/bin/activate`"
+        )
+        return env_dir
+
+
+class CondaBuilder(EnvBuilder):
+
+    type: ClassVar = "conda"
+
+    python_version: str = "3"
```
mike0sv, line 107: probably better to use current python version as default?

mlem/contrib/venv.py
```diff
@@ -0,0 +1,173 @@
...
+    """Whether to install in the current conda env"""
+
+    def create_virtual_env(self):
+        ret = None
+        create_cmd = ["--prefix", self.target, f"python={self.python_version}"]
+        try:
+            ret = subprocess.run(
+                ["conda", "create", "-y", *create_cmd], check=True
+            )
+        except FileNotFoundError:
```
mike0sv, line 119: should we just `os.makedirs` instead?

mlem/contrib/venv.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+class CondaBuilder(EnvBuilder):
+
+    type: ClassVar = "conda"
+
+    python_version: str = "3"
+    """The python version to use"""
+    current_env: Optional[bool] = False
+    """Whether to install in the current conda env"""
```
mike0sv, line 110: `conda_channels` is a good option to add

mlem/contrib/venv.py
```diff
@@ -0,0 +1,173 @@
...
+            assert self.target is not None
+            ret = self.create_virtual_env()
+            assert ret == 0
+            env_dir = self.target
+            env_exe = os.path.join(env_dir, "bin", "python")
+        try:
+            if conda_based_packages:
+                ret = subprocess.run(
+                    [
+                        "conda",
```
mike0sv, line 150: probably should check that `conda` cmd is available

mlem/contrib/venv.py
madhur-tandon: I did it on a package level

mlem/core/requirements.py
```diff
@@ -293,6 +293,26 @@ class UnixPackageRequirement(Requirement):
     package_name: str
 
 
+class CondaPackageRequirement(Requirement):
```
mike0sv, line 296: they can be added only manually for now?

mike0sv (commented):
If you want to go with OOP for venv/conda, go full OOP: signatures of abstract methods should be fixed. Also `build` method probably should be common and just use abstract methods where needed. Otherwise OOP is not really needed

mlem/contrib/venv.py
madhur-tandon: For places where we use `subprocess.call`, we don't know what errors can occur so the question is -- which ones do we handle? Or should we just handle a general `Exception` -- but that will be too broad I guess.

mlem/contrib/venv.py
madhur-tandon: I don't think so, because `conda` create does more than just making a directory.

mlem/core/requirements.py
madhur-tandon: Yes, How can we determine this automatically etc.? Could use some sense of direction in this.

mlem/contrib/requirements.py
madhur-tandon: done

mlem/contrib/venv.py
madhur-tandon: done

mlem/contrib/venv.py
madhur-tandon: done, handled 3 kinds -- `FileNotFoundError`, `subprocess.CalledProcessError` and `subprocess.TimeoutExpired`

mlem/contrib/venv.py
madhur-tandon: done

mlem/contrib/venv.py
madhur-tandon: done

madhur-tandon:
> Also `build` method probably should be common.

Should it be really? I say since each package manager has a different mechanism of installing things really. I still vote for being separate. So `build` is different for `conda` and `venv` really.

mlem/contrib/venv.py
```diff
@@ -0,0 +1,173 @@
...
+class EnvBuilder(MlemBuilder):
+    """MlemBuilder implementation for building virtual environments"""
+
+    type: ClassVar = "env"
+
+    target: Optional[str] = "venv"
+    """Name of the virtual environment"""
+
+    @abstractmethod
+    def create_virtual_env(self):
```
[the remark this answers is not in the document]
madhur-tandon: now it returns nothing, so it is consistent for the 2 classes now

mlem/contrib/requirements.py
```diff
@@ -0,0 +1,37 @@
...
+from mlem.core.objects import MlemBuilder, MlemModel
+from mlem.core.requirements import Requirements
+from mlem.ui import EMOJI_OK, EMOJI_PACK, echo
+
+REQUIREMENTS = "requirements.txt"
+
+logger = logging.getLogger(__name__)
+
+
+class RequirementsBuilder(MlemBuilder):
```
aguschin, line 14: What happens if requirements have something that should be installed with `apt-get`? Are you going to install it?

I assume we should, but on the other hand this may lead to re-installing and breaking (since `apt-get` / `brew` is not isolated). How should we handle this?

aguschin:
Would be great to see documentation to this to review this starting from testing it out in CLI) @madhur-tandon

madhur-tandon:
> Would be great to see documentation to this

Yeah sure, will open something on top of the `new-doc-structure` branch.

mlem/contrib/requirements.py
madhur-tandon: Yes, this is a question that I am also not sure about. I think this might be going too far I guess. We cannot and we shouldn't be a tool to do every possible thing IMO. Maybe we shouldn't touch OS based dependencies. And it will only complicate stuff. I think `venv` and `conda` are at least good to start with for this PR though.

One more related angle is -- how to determine conda based packages automatically given a model. In the test I wrote, I added it manually for example.

mlem/core/requirements.py
aguschin: WDYM by manually? Also, why this is in `core`, while other parts are in `contrib`? Looks to me that `venv` is more "core" than `conda` TBH. Am I missing something?

mlem/contrib/requirements.py
aguschin: So, we don't install OS-based requirements with this? Let's add a short note to cmd description if possible, and cover this in docs. And print a warning after the command execution, if there are some OS deps for the model (if there are none, no reason to print this warning).

I would assume having something like `--os-deps` flag that will install them could be good. But, we can keep it a separate ticket for now.

mike0sv: We can add support for this for unix only, so that you can use it to install in docker build for example. Probably it's better not to call `apt-get` directly, but instead allow to print requirements of specific type, so you can use it like this `apt-get install -y $(mlem build requirements -m model --type unix)`

mlem/contrib/venv.py
```diff
@@ -0,0 +1,228 @@
...
+        raise NotImplementedError
+
+    @abstractmethod
+    def get_installed_packages(self, env_dir):
```
mike0sv, line 28: `: str`

mlem/contrib/venv.py
```diff
@@ -0,0 +1,228 @@
...
+        raise NotImplementedError
+
+    @abstractmethod
+    def build(self, obj: MlemModel) -> str:
```
mike0sv, line 32: this already present in `MlemBuilder`

mlem/contrib/venv.py
```diff
@@ -0,0 +1,228 @@
...
+        raise NotImplementedError
+
+
+class VenvBuilder(EnvBuilder):
+    class Config:
+        arbitrary_types_allowed = True
+
+    type: ClassVar = "venv"
+
+    no_cache: Optional[bool] = False
```
mike0sv, line 42: Why it is `Optional`? Can it be `None`?

mlem/contrib/venv.py
```diff
@@ -0,0 +1,228 @@
...
+    """Disable cache"""
+    current_env: Optional[bool] = False
```
mike0sv, line 44: Same here

mlem/contrib/venv.py
```diff
@@ -0,0 +1,228 @@
...
+        env_dir = os.path.abspath(self.target)
+
+        self.context = env_spec.ensure_directories(env_dir)
+
+        true_system_site_packages = env_spec.system_site_packages
+        env_spec.system_site_packages = False
+        env_spec.create_configuration(self.context)
+        env_spec.setup_python(self.context)
+        if env_spec.with_pip:
+            env_spec._setup_pip(  # pylint: disable=protected-access
```
mike0sv, line 60: It seems you always create spec with `with_pip=True`. And why you need protected access?

mlem/contrib/venv.py
```diff
@@ -0,0 +1,228 @@
...
+    def build(self, obj: MlemModel):
+        if self.current_env:
+            if (
+                os.getenv("VIRTUAL_ENV") is None
+                or sys.prefix == sys.base_prefix
+            ):
+                raise MlemError("No virtual environment detected.")
+            echo(EMOJI_PACK + f"Detected the virtual env {sys.prefix}")
+            env_dir = sys.prefix
+            if platform.system() == "Windows":
```
mike0sv, line 98: seems like a good candidate for a separate method

mlem/contrib/venv.py
```diff
@@ -0,0 +1,228 @@
...
+        echo(EMOJI_PACK + "Installing the required packages...")
+        # Based on recommendation given in https://pip.pypa.io/en/latest/user_guide/#using-pip-from-your-program
+        install_cmd = [env_exe, "-m", "pip", "install"]
+        if self.no_cache:
+            install_cmd.append("--no-cache-dir")
+        install_cmd.extend(obj.requirements.to_pip())
+        try:
+            subprocess.run(install_cmd, check=True)
+        except FileNotFoundError as e:
+            raise MlemError(f"Executable for pip not found.\n{e}") from e
```
mike0sv, line 118: You still print the actual error in the end and you have multiple lines like this. You can extract error handling in a separate method, or catch all the exceptions at once and raise single exception like `error running pip: {e}`

mlem/contrib/venv.py
```diff
@@ -0,0 +1,228 @@
...
+        except subprocess.CalledProcessError as e:
+            raise MlemError(f"Couldn't install packages from pip.\n{e}") from e
+        except subprocess.TimeoutExpired as e:
+            raise MlemError(f"Installing pip packages timed out.\n{e}") from e
+        echo(
+            EMOJI_OK
+            + f"virtual environment `{self.target}` is ready, activate with `source {self.target}/bin/activate`"
```
mike0sv, line 125: I doubt this will work for windows

mlem/contrib/venv.py
mike0sv: You can also add error msg as a parameter for the method

mlem/contrib/venv.py
madhur-tandon: because `_setup_pip` is a protected function.
And we need pip in the new environment so as to use it to install packages with it.

mlem/core/requirements.py
mike0sv: @aguschin: manually means our `get_requirements` method will never yield any `CondaPackageRequirement` and only way to have this in your metadata is to do smth like `model.requirements.add(...); model.update()`
And yes, it can be moved to `mlem.contrib.conda` and should work too (esp since there are no additional requirements for this extension).
@madhur-tandon: can give me an example of this conda requirements?
I can imagine 2 cases: 
- those are just python pip requirements, but you want to install them through conda. In this case you dont need a separate requirement implementation. Instead you would change CondaBuilder logic to try and install regular pip requirements with conda. Maybe you can check if those exist with something like `conda find` (not sure about it)
- There are some requirements that can be installed only through conda. In this case, can we determine them automatically? Probably not, since if it's known that eg `xgboost` package needs `libgomp1` package, it will have it as a dependency in conda ecosystem and we dont need to add this. I might be wrong here

mike0sv: You can also add a field to CondaBuilder with list of additional CondaRequirements to install

mlem/core/requirements.py
```diff
@@ -337,6 +364,22 @@ def modules(self) -> List[str]:
     def expanded(self) -> "Requirements":
         return expand_requirements(self)
 
+    def _add_conda(self, requirement: CondaPackageRequirement):
```
mike0sv, line 367: I guess that is the reason to put it here. We probably should move this logic into `Requirement` class so that requirements check their compatibility themselves

mike0sv: But that's probably for another issue

mlem/contrib/venv.py
madhur-tandon: done

mlem/contrib/venv.py
madhur-tandon: removed and done

mlem/contrib/venv.py
madhur-tandon: yeah, I should remove `Optional`, done!

mlem/contrib/venv.py
madhur-tandon: removed `Optional` here as well

mlem/contrib/venv.py
```diff
@@ -0,0 +1,228 @@
...
+    class Config:
+        arbitrary_types_allowed = True
+
+    type: ClassVar = "venv"
+
+    no_cache: Optional[bool] = False
+    """Disable cache"""
+    current_env: Optional[bool] = False
+    """Whether to install in the current virtual env, must be active"""
+    context: Optional[SimpleNamespace] = None
```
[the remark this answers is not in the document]
madhur-tandon: done

mlem/contrib/venv.py
madhur-tandon: done

mlem/contrib/venv.py
madhur-tandon: done

mlem/core/requirements.py
madhur-tandon: Let's do this later then

mlem/contrib/requirements.py
madhur-tandon: I have done the `unix` thingie and added a test for the same. But the argument is `platform` instead of `type` since `type` is already a `ClassVar`

mlem/core/requirements.py
madhur-tandon: @mike0sv --
1. You shouldn't install packages that are meant to be installed via pip using conda. It is possible that some version is available on pip and not in conda. So we really shouldn't attempt to mix / just use conda instead of pip.

2. Yes, there are some reqs (usually non pythonic) which are only available through conda. I am not aware of how to determine those automatically. But of course, if we know beforehand -- we can add them explicitly, just like `UnixRequirements` are done in the `get_requirements` function currently.

aguschin: The decision is up to you guys, but it looks to me like we're adding some specific functionality with handling conda that is
1. Not that needed - if you want to spin up a new env for the model, venv should be enough. `conda` is something for Data Analysts / ML Developers - when you want to update your env from time to time, clean it, work with it for a long time, etc. When you just want to `serve` or `load` the model, it looks to me more like a temporary and separate env that will serve this purpose only. So `venv` is enough.
2.  Work when you do it manually and looks different from what we do with MLEM, by trying to do as many things automatically for the user as we could.

Consider this, I would suggest extracting conda functionality to a separate PR and leave it alone until we have a clear user request for this. Otherwise we're just wasting time and effort here, trying to come up with a decent solution for a user.

madhur-tandon:
> Would be great to see documentation to this to review this starting from testing it out in CLI

@aguschin here they are: https://github.com/iterative/mlem.ai/pull/191

mlem/contrib/requirements.py
```diff
@@ -0,0 +1,45 @@
...
+    """MlemBuilder implementation for building requirements"""
+
+    type: ClassVar = "requirements"
+
+    target: Optional[str] = None
+    """Target path for the requirements.txt file"""
+    platform: str = "pip"
```
mike0sv, line 21: add a validator to ensure that this is in `list_implementations(Requirement)`

mlem/contrib/requirements.py
```diff
@@ -0,0 +1,45 @@
...
+            with open(os.path.join(self.target), "w", encoding="utf8") as req:
+                logger.debug(
+                    "Auto-determined requirements for model: %s.",
+                    requirements.to_pip(),
+                )
+                req.write(requirement_string + "\n")
+                echo(EMOJI_OK + f"{self.target} generated!")
+
+    def build(self, obj: MlemModel):
+        if self.platform == "unix":
```
mike0sv, line 39: Use `load_impl_ext("requirement", self.platform)` to get python type and then use `.of_type` method to get the list of requirements of this type. If target is `None`, print them, otherwise write them to file.
This way it will work for any requirement type

mike0sv: You may also want to add abstract method to `Requirement` to get "installable" representation of those so you can `" ".join(r.get_repr() for r in reqs)`

mlem/contrib/requirements.py
mike0sv: also, platform is not good choice because there are `file` and `custom` requirements. You can use `req_type` or smth like this

mlem/contrib/venv.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+def get_python_exe_in_virtual_env(env_dir: str, env_type: str = "pip"):
+    if platform.system() == "Windows":
+        if env_type == "pip":
+            env_exe = os.path.join(env_dir, "Scripts", "python.exe")
+        elif env_type == "conda":
+            env_exe = os.path.join(env_dir, "python.exe")
+    else:
+        env_exe = os.path.join(env_dir, "bin", "python")
```
mike0sv, line 22: you can just immediately return instead of setting variable

mlem/contrib/venv.py
```diff
@@ -0,0 +1,187 @@
...
+from typing import ClassVar, List, Optional
+
+from mlem.core.errors import MlemError
+from mlem.core.objects import MlemBuilder, MlemModel
+from mlem.ui import EMOJI_OK, EMOJI_PACK, echo
+
+
+def get_python_exe_in_virtual_env(env_dir: str, env_type: str = "pip"):
+    if platform.system() == "Windows":
+        if env_type == "pip":
```
mike0sv, line 17: This probably should be a boolean flag and not a string

mlem/contrib/venv.py
mike0sv: Yes, but it is protected for some reason. That means you are not using it how it is expected. You should investigate why is it protected and either find another way to use it (maybe there is a public wrapper or something) or acknowledge that you are "trespassing" and have a good reason for that as well as understanding how it can break and be ready for this

mlem/contrib/venv.py
```diff
@@ -0,0 +1,187 @@
...
+                or sys.prefix == sys.base_prefix
+            ):
+                raise MlemError("No virtual environment detected.")
+            echo(EMOJI_PACK + f"Detected the virtual env {sys.prefix}")
+            env_dir = sys.prefix
+            env_exe = get_python_exe_in_virtual_env(env_dir)
+        else:
+            echo(EMOJI_PACK + f"Creating virtual env {self.target}...")
+            self.create_virtual_env()
+            assert self.context is not None
```
mike0sv, line 106: since `context` field is for cache, it makes sense to create accessor method like `get_context` that will check run `create_virtual_env` if it is empty and always return a context.

mlem/core/requirements.py
mike0sv: Ok we can leave it as it is for now here - uses will have to add those manually. We can move conda req implementation to `contrib` and add an optional list of additional conda reqs for builder - that should suffice. Also create an issue to implement automagic conda detection for later

mlem/contrib/venv.py
madhur-tandon: done, using `use_conda_env` as True or False now

mlem/contrib/venv.py
madhur-tandon: yes, makes sense, this means we won't need to call `create_virtual_env` directly now for creating the env since `get_context` will handle that for us.

mlem/contrib/requirements.py
madhur-tandon: changed to `req_type` and added a validator that raises a `ValueError`

mlem/contrib/requirements.py
madhur-tandon: done, `custom` requirements is left I guess, the last comment of yours basically.

mlem/core/requirements.py
madhur-tandon: ok so now I have moved it to `contrib` and it is now a list for `CondaBuilder` (with default as `[]` i.e. an empty list) aka it can be supplied by the user via command line I guess. This avoids playing around / modifying / or accessing the `requirements` attribute i.e. the user won't have to do `model_meta.requirements.add()`. So, with that, we don't have the other helper methods such as `self.of_type(CondaPackageRequirement)` or `_add_conda` since the user supplies a list directly to the builder. CC: @mike0sv @aguschin 

I hope this is the correct interpretation / conclusion of the discussion.
Otherwise I can revert the last commit.

mlem/contrib/venv.py
madhur-tandon: actually, I have removed the usage of `context` altogether, so now, we don't have to deal with it + there is no issue of `protected access` as well.

mlem/contrib/venv.py
madhur-tandon: removed usage of `context` altogether so we don't have the issue of `protected access` now

mlem/contrib/venv.py
```diff
@@ -0,0 +1,228 @@
...
+        env_spec.setup_python(self.context)
+        if env_spec.with_pip:
+            env_spec._setup_pip(  # pylint: disable=protected-access
+                self.context
+            )
+        if not env_spec.upgrade:
+            env_spec.setup_scripts(self.context)
+            env_spec.post_setup(self.context)
+        if true_system_site_packages:
+            env_spec.system_site_packages = True
```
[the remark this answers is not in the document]
madhur-tandon: all this was complex and is removed, we no longer use `context` and other associated internal aspects.

mlem/contrib/requirements.py
madhur-tandon: hmm, one does need a identifier on when to materialize vs. when to print and a check on that identifier is basically an if/else thing so we can't really avoid it I guess. Do you have any suggestions @mike0sv ?

mlem/contrib/requirements.py
```diff
@@ -0,0 +1,71 @@
...
+from mlem.core.requirements import Requirement
+from mlem.ui import EMOJI_OK, EMOJI_PACK, echo
+from mlem.utils.entrypoints import list_implementations
+
+REQUIREMENTS = "requirements.txt"
+
+logger = logging.getLogger(__name__)
+
+
+class CondaPackageRequirement(Requirement):
```
mike0sv, line 18: Probably should be near `CondaBuilder`

madhur-tandon: will move to `mlem/contrib/venv.py`

madhur-tandon: done

mlem/contrib/requirements.py
```diff
@@ -0,0 +1,52 @@
...
+            )
+        return req_type
+
+    def build(self, obj: MlemModel):
+        req_type_cls = load_impl_ext(Requirement.abs_name, self.req_type)
+        assert issubclass(req_type_cls, Requirement)
+        reqs = obj.requirements.of_type(req_type_cls)
+        if self.target is None:
+            reqs_representation = [r.get_repr() for r in reqs]
+            requirement_string = "\n".join(reqs_representation)
```
mike0sv, line 47: Maybe `" ".join`? Will it work like `pip install $(mlem build requirements -m model)` RN?

tests/contrib/test_requirements.py
```diff
@@ -0,0 +1,45 @@
...
+    builder = RequirementsBuilder(target=path)
+    builder.build(model_meta)
+    with open(path, "r", encoding="utf-8") as f:
+        assert model_meta.requirements.to_pip() == f.read().splitlines()
+
+
+def test_build_reqs_with_invalid_req_type():
+    with pytest.raises(ValidationError) as exc:
+        RequirementsBuilder(req_type="invalid")
+    assert "req_type invalid is not valid." in str(exc.value)
```
mike0sv, line 21: use `match` in `pytest.raises`

tests/contrib/test_venv.py
```diff
@@ -0,0 +1,84 @@
...
+
+def test_build_venv(tmp_path, model_meta):
+    path = str(tmp_path / "venv")
+    builder = VenvBuilder(target=path)
+    env_dir = builder.build(model_meta)
+    installed_pkgs = (
+        builder.get_installed_packages(env_dir).decode().splitlines()
+    )
+    for each_req in model_meta.requirements.to_pip():
+        assert each_req in installed_pkgs
```
mike0sv, line 57: create sets and compare them once instead?

tests/contrib/test_venv.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+def test_install_in_current_venv_not_active(tmp_path, model_meta):
+    path = str(tmp_path / "venv")
+    builder = VenvBuilder(target=path, current_env=True)
+    with pytest.raises(MlemError) as e:
+        builder.build(model_meta)
+    assert "No virtual environment detected" in str(e.value)
```
mike0sv, line 65: `match`

tests/contrib/test_venv.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+def test_install_in_current_active_venv(tmp_path, model_meta):
+    old_sys_prefix = sys.prefix
```
mike0sv, line 69: better to do it in a fixture (so you can restore it even if test fails)

mlem/core/requirements.py
```diff
@@ -425,21 +480,13 @@ def new(cls, requirements: "AnyRequirements" = None):
             return Requirements(__root__=[])
         return resolve_requirements(requirements)
 
-    def materialize_custom(self, path: str):
```
mike0sv, line 428: lets keep it, just delegate to `CustomRequirement.materialize`. It will allow omit unnecessary imports of `CustomRequirement`

mlem/contrib/requirements.py
madhur-tandon: I have replaced it with `" "`, but `pip install $(mlem build requirements -m model)` doesn't work because of the first line with the emoji i.e. `⏳️ Loading model from sk-model.mlem`.

Is there a way to ignore all this `echo` related stuff?

mlem/core/requirements.py
```diff
@@ -273,6 +308,13 @@ class FileRequirement(CustomRequirement):
     module: str = ""
     """Ignored"""
 
+    def get_repr(self):
+        raise NotImplementedError
+
+    @classmethod
+    def materialize(cls, reqs, target: str):
```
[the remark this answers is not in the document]
madhur-tandon: removed

tests/contrib/test_requirements.py
madhur-tandon: done

tests/contrib/test_venv.py
madhur-tandon: done

tests/contrib/test_venv.py
madhur-tandon: done

tests/contrib/test_venv.py
madhur-tandon: done

mlem/core/requirements.py
madhur-tandon: done

mlem/contrib/requirements.py
aguschin: Good question! This is a good example when we should try our tools in a user scenario to find out how it will work in fact :)

This is what I heard multiple times from Dmitry and Ivan, btw. See this https://github.com/iterative/mlem/issues/390 Let's discuss it there, so it won't get lost.

mike0sv: We can add `--quiet/-q` to `mlem-callback` that will disable standard cli output

mike0sv (approved):

mlem/contrib/requirements.py
madhur-tandon: Is this to be done? As a separate PR maybe? Or I need to add this `--quiet` stuff somewhere? I do see stuff related to `mlem_group_callback` so wondering if it is to be accommodated there?

Also, should this PR be merged since it's approved OR should we do the above first?

PR #437, opened 2022-10-11
allow to pass pathlib based Paths in save and load APIs

### LINKED ISSUE #428: accomodate pathlib objects for paths in `mlem.api`
context https://github.com/iterative/mlem/issues/359#issuecomment-1265622505

madhur-tandon:
Closes https://github.com/iterative/mlem/issues/428

### DIFF
--- a/mlem/core/metadata.py
+++ b/mlem/core/metadata.py
@@ -3,7 +3,7 @@ Functions to work with metadata: saving, loading,
 searching for MLEM object by given path.
 """
 import logging
-import pathlib
+import os
 import posixpath
 from typing import Any, Dict, Optional, Type, TypeVar, Union, overload
 
@@ -44,7 +44,7 @@ def get_object_metadata(
 
 def save(
     obj: Any,
-    path: Union[str, pathlib.Path],
+    path: Union[str, os.PathLike],
     project: Optional[str] = None,
     sample_data=None,
     fs: Optional[AbstractFileSystem] = None,
@@ -75,13 +75,13 @@ def save(
         sample_data,
         params=params,
     )
-    path = str(path) if isinstance(path, pathlib.Path) else path
+    path = os.fspath(path)
     meta.dump(path, fs=fs, project=project, index=index, external=external)
     return meta
 
 
 def load(
-    path: Union[str, pathlib.Path],
+    path: Union[str, os.PathLike],
     project: Optional[str] = None,
     rev: Optional[str] = None,
     batch_size: Optional[int] = None,
@@ -99,7 +99,7 @@ def load(
     Returns:
         Any: Python object saved by MLEM
     """
-    path = str(path) if isinstance(path, pathlib.Path) else path
+    path = os.fspath(path)
     meta = load_meta(
         path,
         project=project,
@@ -117,7 +117,7 @@ T = TypeVar("T", bound=MlemObject)
 
 @overload
 def load_meta(
-    path: Union[str, pathlib.Path],
+    path: Union[str, os.PathLike],
     project: Optional[str] = None,
     rev: Optional[str] = None,
     follow_links: bool = True,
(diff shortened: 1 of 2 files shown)


mlem/core/metadata.py
```diff
@@ -43,7 +44,7 @@ def get_object_metadata(
 
 def save(
     obj: Any,
-    path: str,
+    path: Union[str, pathlib.Path],
```
skshetry, line 47: Consider using [`os.PathLike`](https://docs.python.org/3/library/os.html#os.PathLike) instead.

mlem/core/metadata.py
```diff
@@ -74,12 +75,13 @@ def save(
         sample_data,
         params=params,
     )
+    path = str(path) if isinstance(path, pathlib.Path) else path
```
skshetry, line 78:
```suggestion
    path = os.fspath(path)
```

See [`os.fspath`](https://docs.python.org/3/library/os.html#os.fspath) and [PEP-0519](https://peps.python.org/pep-0519).

mlem/core/metadata.py
madhur-tandon: Aah, thanks @skshetry

mlem/core/metadata.py
madhur-tandon: Thanks again

madhur-tandon:
@skshetry Unfortunately the above suggestion regarding `os.fspath` and `os.PathLike` leads to a lot of issues (around 35 of them) with `mypy` which is weird since one can pass `str` to variables defined with `os.PathLike`.

No other issues, tests are unaffected. So keeping this PR as it is for now.

skshetry:
> @skshetry Unfortunately the above suggestion regarding `os.fspath` and `os.PathLike` leads to a lot of issues (around 35 of them) with `mypy` which is weird since one can pass `str` to variables defined with `os.PathLike`.

I meant to only replace pathlib.Path with os.PathLike, not the union typehint: `Union[str, os.PathLike]`.

tests/core/test_metadata.py
```diff
@@ -39,6 +40,20 @@ def test_saving_with_project(model, tmpdir):
     load_meta(path)
 
 
+def test_saving_with_pathlib(model, tmpdir):
```
skshetry, line 43: Unrelated to this PR, but use of `tmpdir` and `tmpdir_factory` is discouraged and `tmp_path`/`tmp_path_factory` is preferred which returns `pathlib.Path` objects.

skshetry: yup, as I mentioned, it's unrelated to your PR and I am not asking for a change. Just pointing out what I noticed.

madhur-tandon: Yeah, thanks. I will CC @aguschin and @mike0sv here to take note as well. Will probably open another issue to take note.

mlem/core/metadata.py
```diff
@@ -43,7 +44,7 @@ def get_object_metadata(
 
 def save(
     obj: Any,
-    path: str,
+    path: Union[str, os.PathLike],
```
aguschin, line 47: I see, this covers `mlem.api.load` and `mlem.api.save`. Should we cover all other methods in `mlem.api`?

madhur-tandon: I don't see a pressing need to do those now, because the main purpose was to pass paths for loading and saving. Other APIs have paths for model, which no one really uses the `pathlib` library for.

If needed, we can tackle in another PR.

aguschin: Sorry, didn't get your argument. What is the different between `mlem.api.load` and `mlem.api.apply` in this case? In both places you pass a path. It could be a string and it could be a pathlib object.

aguschin: ok, sure! Then I guess this issue is done, I'll open another one if needed.

aguschin (approved):

PR #446, opened 2022-10-20
use gitlabfs and bitbucketfs from morefs

skshetry:


skshetry:
Is there a way to write these tests in a high-level way, without using `GitlabFileSystem`?
https://github.com/iterative/mlem/blob/86105ee25b99e15ed6f0259ab556966277a0aad1/tests/contrib/test_gitlab.py#L16-L25

setup.py
```diff
@@ -19,7 +19,7 @@
     "cached-property",
     "entrypoints",
     "gitpython",
-    "python-gitlab",
+    "morefs[gitlab] @ git+https://github.com/skshetry/morefs@gitlabfs",
```
aguschin, line 22: Ok, I guess you need to merge this first https://github.com/iterative/morefs/pull/15, and then fix it like
```suggestion
    "morefs",
```
Could you please clarify the syntax here? Wonder how it works. It's going to be `pip install git+https://github.com/skshetry/morefs@gitlabfs`, but why `morefs[gitlab]` and why `[gitlab]` specifically?

aguschin (approved):

aguschin:
Thanks @skshetry. Looks like you can't trigger the CI by youself, right? I wonder because `@all` iterative team has `write` permissions on the repo.

setup.py
skshetry, line 22:
```suggestion
    "morefs[gitlab] @ git+https://github.com/iterative/morefs@gitlabfs",
```

skshetry: This is temporary, I'll merge that PR after all the tests pass here (testing gitlabfs and bitbucketfs is almost impossible).

> why morefs[gitlab] and why [gitlab] specifically?

`morefs` provides multiple filesystems, each with their own required dependencies. These are `extra-dependencies`, `morefs[gitlab]` will install required dependencies for gitlabfs (i.e. `python-gitlab`), and `morefs[bitbucket]` will install `requests` for example.

skshetry:
> Thanks @skshetry. Looks like you can't trigger the CI by youself, right? I wonder because `@all` iterative team has `write` permissions on the repo.

yeah, I don't have permission to trigger CI.

setup.py
aguschin: Ah, got it! 😂 thanks for the patience in the explanation)

aguschin:
added you as a maintainer then!

mlem/contrib/gitlabfs.py
```diff
@@ -176,17 +27,12 @@ def get_kwargs(cls, uri):
...
 
     @classmethod
     def check_rev(cls, options):
-        gl = gitlab.Gitlab()
-        try:
-            gl.projects.get(options["project_id"]).branches.get(options["sha"])
-            return True
-        except GitlabGetError:
-            return False
+        return check_rev(options["project_id"]).branches.get(options["sha"])
```
skshetry, line 35:
```suggestion
        return check_rev(options["project_id"], options["sha"])
```

skshetry:
Still, I cannot approve the workflow. :(

aguschin:
> Still, I cannot approve the workflow. :(

🤷🏻‍♂️
ok, I'll be approving the workflow then!

aguschin:
@skshetry, could you please set this to be merged in `release/0.3.0`? And apply your commits on top of that PR. Tests for `main` branch will fail now, since `mlem-test` repos are updated to match `release/0.3.0` (there are breaking changes now).

skshetry:
I am off next-week, so keeping this as a draft. One thing that is preventing from migrating BitbucketFileSystem is this config which it is using. I am not entirely sure how it's used and how it can be avoided.

https://github.com/iterative/mlem/blob/86105ee25b99e15ed6f0259ab556966277a0aad1/mlem/contrib/bitbucketfs.py#L86

skshetry:
Also, we have dropped support for Python3.7 in all dvc projects. I am okay with adding support for 3.7 in `morefs`, but what do you think about dropping support in mlem/gto? Python 3.7 EOL is still 8 months away though.

aguschin:
Oh, how fast the time runs! I'm ok with dropping 3.7 support in the new release. @mike0sv WDYT?

mike0sv:
#455

mike0sv:
As for bitbucket config, since you can't authenticate via url, we use env variables to do that. Think `AWS_SECRET_KEY` for `s3://...`
`BitbucketConfig` is pydantic's BaseSettings, so it loads values from env variables (as well as mlem config)
I'm not sure why there is no similar thing for gitlab. It seems that `token` and `username` fields are not  used at all, but thay should be (otherwise it will not work for private repos)

aguschin:
> @skshetry, could you please set this to be merged in `release/0.3.0`? And apply your commits on top of that PR.

UPD: we just merged `release/0.3.0` in `main`, so now this should be applied on top of `main` again

aguschin:
> Also, we have dropped support for Python3.7 in all dvc projects. I am okay with adding support for 3.7 in morefs, but what do you think about dropping support in mlem/gto? Python 3.7 EOL is still 8 months away though.

So, @mike0sv do you think we should drop 3.7 support or not? I kept the CI for 3.7 for now in #455. Btw, would be good if analytics could tell us the Python version used https://github.com/iterative/telemetry-python/issues/47

skshetry:
@aguschin, https://pypistats.org/packages/mlem.

mike0sv:
Let's drop it. It's not like it will immediately stop working anyway

skshetry:
I haven't been able to work on this. Will try to pick it up next week. Last time, the issue that I was battling with was `BitbucketConfig` on how `_get_kwargs_from_urls/_strip_protocol` requires that. I'd only like to provide API for passing configs through the constructor (and avoid `env` for now just like GitHubFileSystem), so any pointers on how to decouple would be awesome.

mike0sv:
The problem is that you have to match parts of the URL with existing branches to understand which part is rev and which is path. Eg for github it looks like this: `https://github.com/org/repo/tree/is/this/still/branch/or/path/already`.
To solve this, we just list all branches and tags and match them against everything after `tree`. So if we have a branch or tag `/is/this/still/branch`, we'll know that its rev and `or/path/already` is path.
The problem here is that if you repo is private, you need credentials to list branches and tags. And `fsspec` interface for parsing kwargs from urls does not provide them as an argument for this classmethod.
In github case fsspec documentation says that you should replace `/` in rev for `%2F`, so it will be `../tree/is%2Fthis%2Fstill%2Fbranch/or/path/already`. This is of course unusable if we want our users to be able to provide urls via cli for example. This was one of the reasons to introduce our URI resolvers on top of fsspec logic.
Here are some ways we can do this.
1. Use envs to get credentials. Probably means that you wont be able to pass them to constructor (it's called after uri resolving, so no way to use whatever you passed there)
2. Do it like github. Move branch matching logic to our uri resolver
3. I had some other option in mind, but I forgot it. But I remember it being worse than other two so I dont want to try and remember it

mike0sv:
interesting side note: you cant create such combination of rev and path so it collides, at least on github. 
Eg if you have two files in repo named `b, a/b`, you wont be able to create branch `main/a`

aguschin:
Hi @skshetry. Do you want to complete this?

aguschin:
I removed CI for 3.7 python, please merge `main` into your branch.

skshetry:
I do want to complete this, but I have more important things on my plate right now. So I'll let this sit in draft, and pickup when I can. Thanks and sorry for no updates.

skshetry:
Closing, I don't think I'll have time to work on this soon. Sorry. :(

Issue #445, opened 2022-10-20
De-duplicate methods in `.mlem` model metafile


aguschin:
If we TAL at metafile, we can see the duplication of methods as `predict` + `sklearn_predict`, `predict_proba` + `sklearn_predict_proba`. The reason to introduce them was to give a common interface to models from different frameworks. E.g. sklearn's `predict` and someotherframework's `get_prediction` would have the same endpoint once deployed.

I think we shouldn't duplicate them here. It should be configurable on the server level. This will make future changes depend not on the time when you saved the model, but on the MLEM version.

```yaml
artifacts:
  data:
    hash: 5a38e5d68b9b9e69e9e894bcc9b8a601
    size: 163651
    uri: rf
model_type:
  methods:
    predict:
      args:
        - name: data
          type_:
            columns:
              - sepal length (cm)
              - sepal width (cm)
              - petal length (cm)
              - petal width (cm)
            dtypes:
              - float64
              - float64
              - float64
              - float64
            index_cols: []
            type: dataframe
      name: predict
      returns:
        dtype: int64
        shape:
          - null
        type: ndarray
    predict_proba:
      args:
        - name: data
          type_:
            columns:
              - sepal length (cm)
              - sepal width (cm)
              - petal length (cm)
              - petal width (cm)
            dtypes:
              - float64
              - float64
              - float64
              - float64
            index_cols: []
            type: dataframe
      name: predict_proba
      returns:
        dtype: float64
        shape:
          - null
          - 3
        type: ndarray
    sklearn_predict:
      args:
        - name: X
          type_:
            columns:
              - sepal length (cm)
              - sepal width (cm)
              - petal length (cm)
              - petal width (cm)
            dtypes:
              - float64
              - float64
              - float64
              - float64
            index_cols: []
            type: dataframe
      name: predict
      returns:
        dtype: int64
        shape:
          - null
        type: ndarray
    sklearn_predict_proba:
      args:
        - name: X
          type_:
            columns:
              - sepal length (cm)
              - sepal width (cm)
              - petal length (cm)
              - petal width (cm)
            dtypes:
              - float64
              - float64
              - float64
              - float64
            index_cols: []
            type: dataframe
      name: predict_proba
      returns:
        dtype: float64
        shape:
          - null
          - 3
        type: ndarray
  type: sklearn
object_type: model
requirements:
  - module: sklearn
    version: 1.1.2
  - module: numpy
    version: 1.22.4
  - module: pandas
    version: 1.5.0
```

[...]

Issue #283, opened 2022-06-03
Chaining multiple stages and persisting them using MLEM


drorata:
(This is a follow up of a call with @aguschin)

In general, ML prediction consists of:

1. Features engineering
1. Features preprocessing
2. Model fitting

Once these steps are completed, predictions can be made. Note that when predicting value for an unseen data point, the features engineering and processing must take place prior to the actual prediction by the fitted model.

Alex outlined something like the following:

Feature engineering script that persists the function as a MLEM object.

```python
# feature-engineer.py
def add_features(df): 
    # do some features magic
    return df

mlem.api.save(add_features, “models/feature-processing”)
```

A training script which  

```python
# train.py
model = Model().fit(X,y)

mlem.api.save(model, “models/rf”)
```

For prediction purposes, something like the following can be used:

```python
# predict.py

fp = mlem.api.load(“models/feature-processing”)
model = mlem.api.load(“models/rf”)

model.predict(fp.transform(dataframe))
```

The question is now how to package this in a single structure. Doing something like:

```
$ mlem build docker models/rf 
```
 won't be aware of the feature engineering step.

Two options raised by Alex:

## Chanining:

```
$ mlem build docker models/feature-processing models/rf
```

## Creating a dedicated function that will be persisted:

```python
# prepare-model-for-deploy.py

fp = mlem.api.load(“models/feature-processing”)
model = mlem.api.load(“models/rf”)

def chain():
  return model.predict(fp.transform(dataframe))

mlem.api.save(chain, “models/chain”)
```

However, the second option is in many ways overlaps with `dvc.yaml`; at the end, the DAG defined by DVC is already targeting this need.

drorata:
@dberenbaum @mike0sv FYI

mike0sv:
The second option has an advantage because it should work right now. 
Though another disadvantage of it is that you will have an extra set of artifacts with basically the same contents.

I'm thinking we should have some other type of mlem object, which can have links to multiple mlem models inside. And this object can be used to configure those chains and then used as a substitute for model objects in methods like `apply/build/serve`

drorata:
I another disadvantage of the 2nd option is that it introduces additional code whose only purpose is the integration with MLEM. It is not really needed for the ML process itself.

[...]

PR #447, opened 2022-10-25
Binary Serializers

mike0sv:
This is the first step to support images in MLEM.
The idea is to add binary serialization to some `DataType`s in addition to regular, dictionary serialization.

## High level overview:
- Now servers can accept files as a request or return them as a response
- User can specify serializer to override default serialization 
- builtin FileSerializer uses the same logic as `save` for writing datasets, but other file serializers can be implemented and even be default in some cases (eg images)

## Implementation details:
 
Note: when I say serialize/serialization etc I also mean de-* at the same time.

### DataType and Serializer

We already had separate classes for data type and serializers, however the were implemented by the same class. I separated those and also slightly changed methods signature: now serialize also accepts a data type object (so we can implement arbitrary serializers like FileSerializer). But to make it easier to transition and use in general, I added a DataTypeSerializer (! it's not a Serializer subclass !) which is just a wrapper for data type + serializer

Most DataTypes has default serialzier (with the same logic they had before)

Finally I also added abstract BinarySerializer which works with bytes and bytestreams. When you need to serialize them you just check if your serializer is binary and go for it.

### Interfaces, Servers and Clients

We have this `interface.json` endpoint that clients can use to get info about data types that server expects. But now we also need to have info about serializers. 
Models have their signatures inside `methods` field, but those don't have any information about serializers (because serializers come to play only once you serve the model). 
That's why I created some additional types that basically the same as `Signature` and `Argument`, but also have a field for `Serializer`. 
And finally we also want all of this to be configurable from server fields (including #445 changes). For that I also added separate types (the difference is that their fields are optional and they have some helper methods to match interface's methods). And special ServerInterface (mostly the same purpose as `ConformedInterface` had) to apply this configuration before we actually serve the interface. In CLI it looks like this 
image

### FastAPI server

Basically now we create different handlers depending on serializer type used (if it's binary, we create handler that accepts files)

image


mike0sv:
Looks something like this 
image

mike0sv:
I added `BinaryDataType` that only have binary serializer and no regular serializer. Using it I was able to spin up server with pytorch model that accepts image file as input. But for that I wrapped torch model with function to do needed pre-processing: converting bytes to tensor first and then model-specific pre-processing. 
In the end this should work more out-of-the-boxy-y, but that problem is closer to #283 so I'll leave it for now.
In this PR I will also add a possibility to return files as a result of prediction.

mike0sv:
To try it out you can use this code
```
from torch import frombuffer, uint8
from torchvision.io import decode_image, read_image
from torchvision.models import ResNet50_Weights, resnet50

import mlem.api
from mlem.api import serve
from mlem.core.metadata import load


def raw_model():
    weights = ResNet50_Weights.DEFAULT
    categories = weights.meta["categories"]
    model = resnet50(weights=weights)
    model.eval()

    preprocess = weights.transforms()

    def model_f(img):
        batch = preprocess(img).unsqueeze(0)
        prediction = model(batch).squeeze(0).softmax(0)
        class_id = prediction.argmax().item()
        category_name = categories[class_id]
        return category_name

    return model_f

def torch_image_input(f):
    def inner(arg):
        tensor = frombuffer(arg, dtype=uint8)
        return f(decode_image(tensor))
    return inner



def main():
    with open("cat.jpg", "rb") as f:
        payload = f.read()
    #
    # img = decode_image(frombuffer(payload, dtype=uint8))
    img = read_image("cat.jpg")

    model = raw_model()

    category_name = model(img)
    print(f"{category_name}")

    mmodel = mlem.api.save(model, "torch_resnet", sample_data=img)

    pred = mlem.api.apply(mmodel, img)
    print(pred)

    mmodel_bin = mlem.api.save(torch_image_input(model), "torch_resnet_bin", sample_data=payload)
    pred = mlem.api.apply(mmodel_bin, payload)
    print(pred)
    load("torch_resnet_bin")

def s():
    serve("torch_resnet_bin", "fastapi")

if __name__ == '__main__':

    main()
    s()
```

And it looks like this
image

mike0sv:
I also separated serializers from data types (previously they often were implemented by the same class).
I will now fix some tests and think how to specify serializers better

mike0sv:
I'm thinking of something like this
```python
serve("model", "fastapi", predict_data_serializer="file", predict_returns_serializer=FileSerializer(format="csv"))
```

mike0sv:
So now all of this works
```python
from mlem.api import serve
from mlem.contrib.fastapi import FastAPIServer
from mlem.core.data_type import FileSerializer


def main():
    serve("sklearn", FastAPIServer(
        request_serializer=FileSerializer(),
        response_serializer=FileSerializer()
    ))

    serve("sklearn", FastAPIServer(
        serializers={"predict": {"X": FileSerializer()},
                     "predict_proba": {"returns": FileSerializer()}}
    ))

    serve("sklearn", "fastapi", **{"serializers.predict.X": "file",
                                   "serializers.predict_proba.returns": "file"})


if __name__ == '__main__':
    main()
```

as well as 
```bash 
$  mlem serve fastapi -m sklearn --response_serializer file
```

mike0sv:
And images can be dealt with like this
```python
from typing import Any, BinaryIO, Iterator, Union

from torch import frombuffer, uint8
from torchvision.io import decode_image, read_image
from torchvision.models import ResNet50_Weights, resnet50

import mlem.api
from mlem.api import serve
from mlem.contrib.torch import TorchTensorDataType
from mlem.core.data_type import BinarySerializer


def raw_model():
    weights = ResNet50_Weights.DEFAULT
    categories = weights.meta["categories"]
    model = resnet50(weights=weights)
    model.eval()

    preprocess = weights.transforms()

    def model_f(img):
        batch = preprocess(img).unsqueeze(0)
        prediction = model(batch).squeeze(0).softmax(0)
        class_id = prediction.argmax().item()
        category_name = categories[class_id]
        return category_name

    return model_f


class TorchImageSerializer(BinarySerializer):
    def serialize(self, data_type: TorchTensorDataType, instance: Any) -> bytes:
        raise NotImplementedError

    def dump(self, data_type: TorchTensorDataType, instance: Any) -> Iterator[
        BinaryIO]:
        raise NotImplementedError

    def deserialize(self, data_type: TorchTensorDataType,
                    obj: Union[bytes, BinaryIO]) -> Any:
        if isinstance(obj, bytes):
            buffer = obj
        else:
            buffer = obj.read()
        return decode_image(frombuffer(buffer, dtype=uint8))


def main():
    img = read_image("cat.jpg")

    model = raw_model()

    category_name = model(img)
    print(f"{category_name}")

    mlem.api.save(model, "torch_resnet", sample_data=img)
    serve("torch_resnet", "fastapi", request_serializer=TorchImageSerializer())
    

if __name__ == '__main__':
    main()
```

mike0sv:
@aguschin @madhur-tandon I guess this is ready for review. Suggestions in the comments are for subsequent issues.
When we finalize the approach I will add more tests and docs.
@omesser @shcheklein you can jump on this one too

setup.py
```diff
@@ -173,33 +173,35 @@
             "k8s_service_type.loadbalancer = mlem.contrib.kubernetes.service:LoadBalancerService",
             "k8s_service_type.nodeport = mlem.contrib.kubernetes.service:NodePortService",
             "data_reader.lightgbm = mlem.contrib.lightgbm:LightGBMDataReader",
+            "serializer.lightgbm = mlem.contrib.lightgbm:LightGBMDataSerializer",
```
aguschin, line 176: When trying to run your examples, I get:
```
RuntimeError: Form data requires "python-multipart" to be installed.
You can install "python-multipart" with:

pip install python-multipart
```

You need to add it to `setup.py`.

mlem/runtime/server.py
```diff
@@ -19,6 +21,14 @@ class Config:
     abs_name: ClassVar[str] = "server"
     env_vars: ClassVar[Optional[Dict[str, str]]] = None
     additional_source_files: ClassVar[Optional[List[str]]] = None
+    request_serializer: Optional[Serializer] = None
```
aguschin, line 24:
```suggestion
    request_serializer: Optional[str, Serializer] = None
```

mlem/runtime/server.py
```diff
@@ -19,6 +21,14 @@ class Config:
...
+    """Serializer to use for all requests"""
+    response_serializer: Optional[Serializer] = None
```
aguschin, line 26:
```suggestion
    response_serializer: Optional[str, Serializer] = None
```

mlem/runtime/server.py
aguschin: Since it works like 
```python
   ...: serve("get-started/sklearn", FastAPIServer(
   ...:         request_serializer="file",
   ...:         response_serializer="file"
   ...: ))
```

mlem/contrib/fastapi.py
```diff
@@ -6,15 +6,18 @@
 import logging
 from collections.abc import Callable
 from types import ModuleType
-from typing import Any, ClassVar, List, Type
+from typing import Any, ClassVar, Dict, List, Optional, Tuple, Type
```
aguschin, line 9: When I run your [Torchvision example with resnet](https://github.com/iterative/mlem/pull/447#issuecomment-1299909225), I have two handlers: `__call__` and `predict`. While the second works, the first always throws errors like:

<details>

```shell
INFO:     127.0.0.1:60666 - "POST /__call__ HTTP/1.1" 500 Internal Server Error
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/Users/aguschin/.local/share/virtualenvs/mlem-Utz6DvOn/lib/python3.10/site-packages/uvicorn/protocols/http/h11_impl.py", line 404, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
  File "/Users/aguschin/.local/share/virtualenvs/mlem-Utz6DvOn/lib/python3.10/site-packages/uvicorn/middleware/proxy_headers.py", line 78, in __call__
    return await self.app(scope, receive, send)
  File "/Users/aguschin/.local/share/virtualenvs/mlem-Utz6DvOn/lib/python3.10/site-packages/fastapi/applications.py", line 270, in __call__
    await super().__call__(scope, receive, send)
  File "/Users/aguschin/.local/share/virtualenvs/mlem-Utz6DvOn/lib/python3.10/site-packages/starlette/applications.py", line 124, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/Users/aguschin/.local/share/virtualenvs/mlem-Utz6DvOn/lib/python3.10/site-packages/starlette/middleware/errors.py", line 184, in __call__
    raise exc
  File "/Users/aguschin/.local/share/virtualenvs/mlem-Utz6DvOn/lib/python3.10/site-packages/starlette/middleware/errors.py", line 162, in __call__
    await self.app(scope, receive, _send)
  File "/Users/aguschin/.local/share/virtualenvs/mlem-Utz6DvOn/lib/python3.10/site-packages/starlette/middleware/exceptions.py", line 75, in __call__
    raise exc
  File "/Users/aguschin/.local/share/virtualenvs/mlem-Utz6DvOn/lib/python3.10/site-packages/starlette/middleware/exceptions.py", line 64, in __call__
    await self.app(scope, receive, sender)
  File "/Users/aguschin/.local/share/virtualenvs/mlem-Utz6DvOn/lib/python3.10/site-packages/fastapi/middleware/asyncexitstack.py", line 21, in __call__
    raise e
  File "/Users/aguschin/.local/share/virtualenvs/mlem-Utz6DvOn/lib/python3.10/site-packages/fastapi/middleware/asyncexitstack.py", line 18, in __call__
    await self.app(scope, receive, send)
  File "/Users/aguschin/.local/share/virtualenvs/mlem-Utz6DvOn/lib/python3.10/site-packages/starlette/routing.py", line 680, in __call__
    await route.handle(scope, receive, send)
  File "/Users/aguschin/.local/share/virtualenvs/mlem-Utz6DvOn/lib/python3.10/site-packages/starlette/routing.py", line 275, in handle
    await self.app(scope, receive, send)
  File "/Users/aguschin/.local/share/virtualenvs/mlem-Utz6DvOn/lib/python3.10/site-packages/starlette/routing.py", line 65, in app
    response = await func(request)
  File "/Users/aguschin/.local/share/virtualenvs/mlem-Utz6DvOn/lib/python3.10/site-packages/fastapi/routing.py", line 231, in app
    raw_response = await run_endpoint_function(
  File "/Users/aguschin/.local/share/virtualenvs/mlem-Utz6DvOn/lib/python3.10/site-packages/fastapi/routing.py", line 162, in run_endpoint_function
    return await run_in_threadpool(dependant.call, **values)
  File "/Users/aguschin/.local/share/virtualenvs/mlem-Utz6DvOn/lib/python3.10/site-packages/starlette/concurrency.py", line 41, in run_in_threadpool
    return await anyio.to_thread.run_sync(func, *args)
  File "/Users/aguschin/.local/share/virtualenvs/mlem-Utz6DvOn/lib/python3.10/site-packages/anyio/to_thread.py", line 31, in run_sync
    return await get_asynclib().run_sync_in_worker_thread(
  File "/Users/aguschin/.local/share/virtualenvs/mlem-Utz6DvOn/lib/python3.10/site-packages/anyio/_backends/_asyncio.py", line 937, in run_sync_in_worker_thread
    return await future
  File "/Users/aguschin/.local/share/virtualenvs/mlem-Utz6DvOn/lib/python3.10/site-packages/anyio/_backends/_asyncio.py", line 867, in run
    result = context.run(func, *args)
  File "/Users/aguschin/Git/iterative/mlem/mlem/contrib/fastapi.py", line 139, in handler
    result = executor(**{arg_name: arg})
  File "/Users/aguschin/Git/iterative/mlem/mlem/runtime/interface.py", line 204, in executor
    return self.model_type.call_method(method_name, *args)
  File "/Users/aguschin/Git/iterative/mlem/mlem/core/model.py", line 278, in call_method
    output_data = self._call_method(signature.name, *input_data)
  File "/Users/aguschin/Git/iterative/mlem/mlem/core/model.py", line 293, in _call_method
    return getattr(self.model, wrapped)(*input_data)
AttributeError: 'function' object has no attribute 'model_f'
```

</details>

PS: I see that there are images that `predict` cannot handle, but the errors are different (like `RuntimeError: The size of tensor a (4) must match the size of tensor b (3) at non-singleton dimension 0`). So this is something wrong with the handler.

BTW, can it be fixed by https://github.com/iterative/mlem/pull/456 by chance?

aguschin:
Thanks @mike0sv, tried it out. I posted few issues, otherwise it looks good to me :) Could you please create issues for the next steps and link them in this epic? https://github.com/iterative/mlem/issues/454

Or just put them in this epic)

aguschin (approved):

mlem/runtime/server.py
mike0sv: Nope, in my example only `FastAPIServer(request_serializer=FileSerializer())` is allowed. It's the same for all other cases when you provide an instance of server/builder/client/etc - it's nested fields should be instances too.
You are probably confused by the ability to provide strings instead of instances for topmost objects. In this case you also should provide kwargs, and those can have nested strings. Thats because  the topmost object is not constructed in this case and those nested fields will be constructed together with it. But if you already have topmost object, this logic would not work

mlem/contrib/fastapi.py
mike0sv: Looks like different issue, can you try it without torch?

mlem/runtime/server.py
aguschin: you can try this - it works
```python
from mlem.api import serve
from mlem.contrib.fastapi import FastAPIServer
from mlem.core.data_type import FileSerializer

serve("sklearn", FastAPIServer(
    request_serializer=FileSerializer(),
    response_serializer=FileSerializer()
))
```
or maybe be I got you wrong?

mlem/contrib/fastapi.py
aguschin: yes, https://github.com/iterative/mlem/issues/475

mlem/runtime/server.py
mike0sv: This works
```python
FastAPIServer(
    request_serializer=FileSerializer(),
    response_serializer=FileSerializer()
)
```
This does not
```python
FastAPIServer(
    request_serializer="file",
    response_serializer="file"
)
```

aguschin: Sorry, that was a typo in my last comment. This works for me:
image

mlem/contrib/xgboost.py
```diff
@@ -116,11 +102,33 @@ def process(cls, obj: xgboost.DMatrix, **kwargs) -> DataType:
...
+class DMatrixSerializer(DataSerializer[DMatrixDataType]):
+    """Serialzier for XGBoost DMatrix"""
+
+    is_default: ClassVar = True
+    data_class: ClassVar = DMatrixDataType
+
+    def get_model(self, data_type, prefix: str = "") -> Type[BaseModel]:
         raise NotImplementedError
 
+    def serialize(
```
omesser, line 117: Minor peeve: Need to implement (inherited `NotImplementedError`)?
Maybe the `NotImplementedError` can just serialize the class name in the error message and than this wouldn't be needed

omesser (approved):
👍 
If you can update the description per current implementation
How are the serializers currently should be used (per your last comments, probably copy-paste)

PR #461, opened 2022-10-28
Error on extra cli arguments

### LINKED ISSUE #459: Raise error on passing wrong options
Rn you can pass any option to `declare` and it won't error out. This is a problem both because people tend to do typos and because they may think pass incorrect parameters excepting them to affect the deployment (e.g. `--model`). We still have mistakes regarding this in docs, and people will wonder why commands work initially, but it breaks later.
```
$ mlem declare env heroku staging
$ mlem declare deployment heroku app \
    --app_name=example-mlem-get-started-app \
    --mlem_is_awesome_arent_it=yes_hooman \
    --env=staging
💾 Saving deployment to app.mlem
```

Maybe other commands with the magic (`build`, `serve`, `deploy`, `apply`) are affected as well. We need to check this.

mike0sv:
I spent 1 hour to find this one
closes #459

### DIFF
--- a/mlem/cli/main.py
+++ b/mlem/cli/main.py
@@ -17,7 +17,7 @@ from typing import (
 import click
 import typer
 from click import Abort, ClickException, Command, HelpFormatter, Parameter
-from click.exceptions import Exit, MissingParameter
+from click.exceptions import Exit, MissingParameter, NoSuchOption
 from pydantic import ValidationError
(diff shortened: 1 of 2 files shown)


aguschin:
Curious why black pre-commit failed in CI. Pushing a fix anyway :)

aguschin (approved):
Looks good!

PR #467, opened 2022-11-01
Dummy PR to check Python 3.10 tests

aguschin:


aguschin:
@mike0sv @madhur-tandon you're awesome - just TAL at tests in 3.10. They just pass! This is a development with future taken into account. Nice!

PR #468, opened 2022-11-02
Raise error if trying to serve model without signature

mike0sv:


madhur-tandon (approved):

tests/api/test_commands.py
```diff
@@ -240,3 +241,12 @@ def test_build_lazy(model_meta, tmp_path):
...
     )
+
+
+def test_serve_no_signature(model):
+    meta = MlemModel.from_obj(model)
+    with pytest.raises(
+        MlemError,
+        match="Cannot create interface from model with uspecified signature. Please re-save it and provide sample_data argument",
+    ):
+        serve(meta, "fastapi")
```
aguschin, line 252: Looks good! I think we should also issue an Exception when you build a Docker Image or try to deploy model somewhere.

mike0sv: I thought about it, but build may occur in environments without model requirements, which means there will be no access to model signature and we wont be able to check it

aguschin: Can't we just check the presence of the signature? Like `if signature is not None`? Even if we don't load the actual content.

aguschin: From the naive perspective: since it exists in `model.mlem`, some magic should do the trick 🤔

aguschin: Ok, we need to find some solution here, because from the user perspective it can be pretty irritating if MLEM allows to build/deploy a model, but later fails to serve it. Let's think about this some more time then.

aguschin: Btw, do I understand correctly, that if it's `UnspecifiedDataType`s, then we don't need any model requirements installed to local signature? Or we still need let's say `PyTorch` if model was a PyTorch one?
What I mean, is that if we try to deserialize signature and we can't because of missing requirements, then it's not `UnspecifiedDataType` and everything is ok.

But we probably we deserialize signature as part of deserializing a model itself, so that will fail before it reaches the signature?

mike0sv: Yes, we need model requirements because data types are nested fields. Also, we need data type requirements, because theoretically only part of all data types can be unspecified

aguschin: > only part of all data types can be unspecified

If that happens to be, will `serve` fail?

Then it looks like the only alternative to checking raw objects is to try to check anyway - if we can't load some dependencies, we're skipping this, but just issue a warning about it.

aguschin: Good point! Then we need to check servers somehow 🤔 
What I'm advocating here for: we should fail as early on as possible. Otherwise it's going to be a bad experience for users :)

mike0sv: I get it. But not all servers/builders require signature to be present. Eg venv builder
A bit ugly, but we can add classvar for servers  which will enable signature checking.
And we can check raw methods objects `type` field

mlem/runtime/interface.py
```diff
@@ -189,6 +189,22 @@ def load(self, uri: str):
...
+        if any(
+            isinstance(a, UnspecifiedDataType)
+            for method in value.methods.values()
+            for a in method.args
+        ) or any(
+            isinstance(method.returns, UnspecifiedDataType)
+            for method in value.methods.values()
+        ):
+            raise MlemError(
+                "Cannot create interface from model with uspecified signature. Please re-save it and provide sample_data argument"
```
aguschin, line 204:
```suggestion
                "Cannot create interface from model with unspecified signature. Please re-save it and provide `sample_data` argument"
```

tests/api/test_commands.py
```diff
@@ -240,3 +241,12 @@ def test_build_lazy(model_meta, tmp_path):
     build(
         "pip", model_meta, target=str(tmp_path / "build"), package_name="lol"
     )
+
+
+def test_serve_no_signature(model):
+    meta = MlemModel.from_obj(model)
+    with pytest.raises(
+        MlemError,
+        match="Cannot create interface from model with uspecified signature. Please re-save it and provide sample_data argument",
```
aguschin, line 250:
```suggestion
        match="Cannot create interface from model with unspecified signature. Please re-save it and provide `sample_data` argument",
```

aguschin (approved):

aguschin:
I guess I should have facilitate this to be merge a week ago, at least :) Let's solve the follow-up problem separately in #488

PR #470, opened 2022-11-06
minor - mlem apply to print results always

omesser:
when applying results it's worth showing 😄 

## Before:
```bash
❯ mlem apply models/rf new_data.csv  -m predict_proba -i --it "pandas[csv]"
⏳️ Importing object from new_data.csv
⏳️ Loading model from models/rf.mlem
🍏 Applying `predict` method...
❯
```

## After
```bash
❯ mlem apply models/rf new_data.csv  -m predict_proba -i --it "pandas[csv]"
⏳️ Importing object from new_data.csv
⏳️ Loading model from models/rf.mlem
🍏 Applying `predict_proba` method...
[[0.47, 0.24, 0.29]]
❯
```

### DIFF
--- a/mlem/cli/apply.py
+++ b/mlem/cli/apply.py
@@ -166,7 +166,7 @@ def _apply_remote(
             output,
             target_project,
         )
-    if output is None and json:
+    if output is None:
         print(
             dumps(
                 DataAnalyzer.analyze(result).get_serializer().serialize(result)
--- a/mlem/cli/apply.py
+++ b/mlem/cli/apply.py
@@ -114,7 +114,7 @@ def apply(
             output=output,
             batch_size=batch_size,
         )
-    if output is None and json:
+    if output is None:
         print(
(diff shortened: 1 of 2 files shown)


aguschin (approved):
Reproduced. This was lost somewhere...

mike0sv:
I think this is the only place  where `json` option is used? If so, either remove the option, or change `and json` to `or json` instead of removing completely.

PS. I remember that I added the `json` option, but I dont remember why I did it like this🤔

omesser:
> I think this is the only place where `json` option is used? If so, either remove the option, or change `and json` to `or json` instead of removing completely.
> 
> PS. I remember that I added the `json` option, but I dont remember why I did it like this🤔

its used in [set_echo](https://github.com/iterative/mlem/blob/main/mlem/cli/apply.py#L88) in various places here, suppressing the human-readable output, which is good imo

PR #472, opened 2022-11-07
Fix `setup.py` classifiers

aguschin:


mike0sv:
Should we also exclude 3.7?

aguschin:
@skshetry, in which projects do we use it?

aguschin:
@mike0sv, excluded 3.7.

skshetry:
@aguschin, all of the packages mentioned in https://github.com/iterative/status#other-projects + `dql`.

aguschin:
Thanks @skshetry! Why don't you use it in DVC?

skshetry:
No strong reason tbh. All of those packages came after we had a good template, and we keep `dvc` in a good enough shape.

There are minor things like `.md` vs `.rst` debate (dvc mixes both, template uses rst), and `src/` based directories that template forces (which we are okay with in dvc).

mike0sv (changes requested):
Change heroku tests to run on 3.8

aguschin:
Fixed. I guess we should consider moving control over another if-else conditions based on OS and Python (I saw some docker tests that are controlled in code) to GHA workflow as well. WDYT @mike0sv?

mike0sv:
Maybe, let's do it in another housekeeping issue

aguschin:
@mike0sv, you can approve and merge.
checked this works in https://github.com/iterative/mlem/pull/491

mike0sv (approved):

PR #473, opened 2022-11-07
Run tests in CI with Python 3.11

aguschin:


aguschin:
Closing since tests don't pass - miniconda distribution is not there yet https://github.com/iterative/mlem/actions/runs/3408003288/jobs/5668178622

PR #482, opened 2022-11-10
enable apply for raw model objects

### LINKED ISSUE #231: Support applying non-MLEM models in `mlem.api.apply`
In get started, the example shows `y_pred = apply("rf", "test_x.csv", method="predict_proba")` and below says to use `import` if your dataset is not saved as a mlem object.

First, I tried passing in various array-like objects instead of the path to a mlem dataset and was happy to find that they worked 🎉:

```
>>> df = load("test_x.csv")
>>> y_pred = apply("rf", df, method="predict_proba")
>>> arr = df.to_numpy()
>>> y_pred = apply("rf", arr, method="predict_proba")
/Users/dave/miniforge3/envs/mlem/lib/python3.10/site-packages/sklearn/base.py:450: UserWarning: X does not have valid feature names, but RandomForestClassifier was fitted with feature names
  warnings.warn(
>>> l = arr.tolist()
>>> y_pred = apply("rf", l, method="predict_proba")
/Users/dave/miniforge3/envs/mlem/lib/python3.10/site-packages/sklearn/base.py:450: UserWarning: X does not have valid feature names, but RandomForestClassifier was fitted with feature names
  warnings.warn(
```

It might be worth documenting this more clearly so that users understand that they can continue to work with in-memory objects and don't have to serialize and deserialize at each step if they don't want to.
 
On the other hand, passing a sklearn classifier as the model didn't work so well:

```
>>> rf = load("rf")
>>> y_pred = apply(rf, "test_x.csv", method="predict_proba")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/dave/miniforge3/envs/mlem/lib/python3.10/site-packages/mlem/api/commands.py", line 80, in apply
    model = get_model_meta(model)
  File "/Users/dave/miniforge3/envs/mlem/lib/python3.10/site-packages/mlem/api/utils.py", line 37, in get_model_meta
    raise InvalidArgumentError(
mlem.core.errors.InvalidArgumentError: The object RandomForestClassifier(n_jobs=2, random_state=42) is neither ModelMeta nor path to it
```

It would be nice to be able to pass the sklearn object to apply. Many times it's already in memory and that's easier and faster than reloading. It also allows for using `mlem.apply` without having to first use `mlem.save` on the objects.

mike0sv:
closes #231

### DIFF
--- a/mlem/api/commands.py
+++ b/mlem/api/commands.py
@@ -47,7 +47,7 @@ from mlem.ui import (
 
 
 def apply(
-    model: Union[str, MlemModel],
+    model: Union[str, MlemModel, Any],
     *data: Union[str, MlemData, Any],
     method: str = None,
     output: str = None,
@@ -71,7 +71,10 @@ def apply(
         If `output=None`, returns results for given data.
             Otherwise returns None.
     """
-    model = get_model_meta(model)
+    if isinstance(model, (str, MlemModel)):
+        model = get_model_meta(model)
+    else:
+        model = MlemModel.from_obj(model)
     w = model.model_type
     try:
         resolved_method = w.resolve_method(method)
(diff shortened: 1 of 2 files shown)


aguschin (approved):
Ok, that was easy! 👍🏻

PR #485, opened 2022-11-10
Model stages/pre/post processors prototype

mike0sv:
relates to https://github.com/iterative/mlem/issues/283

From UX perspective this should look smth like
```python
from mlem.api import save
from mlem.core.objects import MlemModel

def func(data):
    return sum(data)


def preproc(data):
    return [int(o) for o in data]


def main():
    data = ["1", "1", "1"]

    model: MlemModel = save(func, "preproc", preprocess=preproc)

    value = model(data)

    print(value) # 3
```

Internally however there are a lot of changes. Basically instead of single `ModelType` object we will have a dict of those and some additional field to describe possible orders of those (eg if model has 2 methods, you might want to use different pre/post processors for those). This also enables to pack multiple models into a single object, so they could be served together for example.
Also there should be a possibility to use separately saved models as a "pipeline" steps to enable users to re-use and version pre/post processors separately.

Also need this to work with serializers in #447 seamlessly (eg using serializer as pre-processing step, eg to apply torch models to images)

This will probably be a very big change, possibly breaking backward compatibility, so targeting 0.4 release with this


mike0sv:
Some updates

Basic usage:
```
model = save(obj, path, sample_data=...)
```
Works the same. Internally `model.processors` will be `{"model": {<model_type data>}}` and `model.call_orders` will be `{"predict/__call__/whatever model have": ("model", "predict/.../...")}`. User-wise should be no changes.

Pre/post processing
```
model = save(obj, path, sample_data=..., pre(post)process=<lambda/function/other model>)
```
Same, but whatever you passed as pre/postprocess will be analyzed like any other model and added to processors as `{"preprocess": <preprocessor model_type>}`. Call order will look like `{"predict": [("preprocess", "__call__"), ("model", "predict")]}`. Will work the same even if model has multiple methods (there will be other call order then)

Different pre/post processing
```
model = save(obj, path, methods_sample_data={"predict": ..., "predict_proba": ...}, pre(post)process={"predict": <lambda/function/other model>})
```
Same as before, but pre/postprocessors will be named "preprocess_{method_name}" and call order will look something like `{"predict": [(preprocess_predict, __call__), ("model", "predict"), ("postprocess_predict", "__call__")], "predict_proba": [...same here...]}`

Advanced usage (model merging)
```
model = MlemModel()

model.add_processor("model1", model1)
model.add_processor("model2", model2)

model.call_order["single_model1"] = [("model1", "predict")]
model.call_order["single_model2"] = [("model1", "predict")]
model.call_order["both"] = [("model1", "predict"), ("model2", "predict")]

predict = model.both(data)
```

Advanced user has control over what processors model has and what their call order. All of this is reflected when serving so user can expose multiple models on one server or expose pre/post processing steps separately.

For now there is no way to use separate already saved mlem model as a processor in new model to be able to save and version all the steps separately. But we can add this later (allow `model.processor` entries to be links, for example). Users will be able to construct this meta-model just before serving/deploying

mike0sv:
I am also testing this for CV usecase with this code, and it even works
```
from torchvision.io import read_image
from torchvision.models import ResNet50_Weights, resnet50

import mlem.api
from image_support_ser import TorchImageSerializer
from mlem.api import serve


def main():
    img = read_image("cat.jpg")

    weights = ResNet50_Weights.DEFAULT
    categories = weights.meta["categories"]
    model = resnet50(weights=weights)
    model.eval()
    preprocess = weights.transforms()

    mlem_model = mlem.api.save(model, "torch_resnet2",
                               preprocess=lambda x: preprocess(x).unsqueeze(0),
                               postprocess=lambda x: categories[
                                   x.squeeze(0).softmax(0).argmax().item()],
                               sample_data=img)

    category_name = mlem_model(img)
    print(f"{category_name}")

    # mlem apply torch_resnet image.png --import_type torch_image
    serve("torch_resnet2", "fastapi", request_serializer=TorchImageSerializer())


if __name__ == '__main__':
    main()
```

`image_support_ser` contents
```
import contextlib
from io import BytesIO
from typing import Any, BinaryIO, ClassVar, Iterator, Union

from torch import frombuffer, uint8
from torchvision.io import decode_image, encode_jpeg
from torchvision.transforms import ToPILImage
from torchvision.utils import make_grid, save_image

from mlem.contrib.torch import TorchTensorDataType
from mlem.core.data_type import BinarySerializer


class TorchImageSerializer(BinarySerializer):
    support_files: ClassVar = True

    def _to_buffer(self, instance):
        buffer = BytesIO()
        ToPILImage()(instance).save(buffer, "JPEG")
        buffer.seek(0)
        return buffer

    def serialize(self, data_type: TorchTensorDataType, instance: Any) -> bytes:
        return self._to_buffer(instance).read()

    @contextlib.contextmanager
    def dump(self, data_type: TorchTensorDataType, instance: Any) -> Iterator[
        BinaryIO]:
        yield self._to_buffer(instance)

    def deserialize(self, data_type: TorchTensorDataType,
                    obj: Union[bytes, BinaryIO]) -> Any:
        if isinstance(obj, bytes):
            buffer = obj
        else:
            buffer = obj.read()
        return decode_image(frombuffer(buffer, dtype=uint8))
```

aguschin (approved):
🚀 LGTM!

PR #489, opened 2022-11-16
Enable rev option for local git repos

mike0sv:


mike0sv:
So tests failed because pygit2 is not installed. I think we can just add it to install_requires, but there is other option to make it extra, so you will be able to use this feature only if pygit2 installed. @aguschin WDYT?

aguschin:
👍🏻 
I think adding it to install_requries is valid, since this is a "built-in" functionality for MLEM (compared to let's say fastapi serving).

aguschin:
Some tests fail @mike0sv

aguschin (approved):

mike0sv:
I dont have access to windows machine RN, will check when I do

mike0sv:
It seems that for now the only way is to xfail those tests for windows because of https://github.com/fsspec/filesystem_spec/issues/1125

aguschin (approved):

PR #493, opened 2022-11-19
Update check-test-release.yml

mike0sv:
To be honest I just brainlessly copy-pasted https://github.com/iterative/shtab/pull/121, @casperdcl did I succeed? 😄

### DIFF
--- a/.github/workflows/check-test-release.yml
+++ b/.github/workflows/check-test-release.yml
@@ -153,12 +153,10 @@ jobs:
         echo ::set-output name=tag::${GITHUB_REF#refs/tags/}
         git log --pretty='format:%d%n- %s%n%b---' $(git tag --sort=v:refname | tail -n2 | head -n1)..HEAD > _CHANGES.md
     - if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
-      uses: softprops/action-gh-release@v1
+      name: Release
+      run: |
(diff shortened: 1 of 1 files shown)


aguschin (approved):

aguschin:
Since we need to make a release, I'll go ahead and merge this. No other way to check it'll work I guess :)

aguschin:
Confirmed: successfully works! https://github.com/iterative/mlem/actions/runs/3512101417

.github/workflows/check-test-release.yml
```diff
@@ -153,12 +153,10 @@ jobs:
         echo ::set-output name=tag::${GITHUB_REF#refs/tags/}
         git log --pretty='format:%d%n- %s%n%b---' $(git tag --sort=v:refname | tail -n2 | head -n1)..HEAD > _CHANGES.md
```
casperdcl, line 154: I believe you no longer need these lines (i.e. can delete the `name: Changelog` step)

aguschin: https://github.com/iterative/mlem/pull/500

casperdcl: it was to extract the changelog from git commits (to help draft release notes).

It used to be in a separate (bash) step because the old way used an action (https://github.com/softprops/action-gh-release) rather than plain bash.

Issue #524, opened 2022-12-12
Finding you need to install `google-cloud-storage ` in `from google.cloud import storage`


aguschin:
In some complex cases MLEM can have problems finding what to install. This happens pretty rarely AFAIK, but please upvote and leave a comment if you hit this problem so we can set a priority to investigate and fix.

Example:

```shell
$ pip install google-cloud-storage
```
(see https://cloud.google.com/storage/docs/reference/libraries)

```py
from mlem.api import save
from google.cloud import storage

def f():
    storage

save(f, "storage")

print(storage.__name__)
```

which prints `google.cloud.storage`

saved metafile:
```yaml
artifacts:
  data.pkl:
    hash: 0c3e2cfa007c4e8e25039367c68a29e4
    size: 277
    uri: storage
model_type:
  io:
    type: pickle
  methods:
    __call__:
      args: []
      name: __call__
      returns:
        type: unspecified
  type: callable
object_type: model
requirements:
- module: google
  package_name: google-api-core
```

last lines should have been something like:

```
- module: google.cloud.storage
  package_name: google_cloud_storage
```

aguschin:
Rn we use this:

```py
>>> from importlib_metadata import packages_distributions
>>> packages_distributions()['box']
['python-box']
```
but in such cases as in the first comment, it breaks:
```py
In [10]: pd = packages_distributions()

In [11]: for k in pd:
    ...:     if len(pd[k]) != 1:
    ...:         print(k, pd[k])
    ...:
azure ['azure-identity', 'azure-core', 'azure-storage-blob', 'azure-datalake-store']
google ['google-api-core', 'googleapis-common-protos', 'google-cloud-core', 'google-cloud-storage', 'google-resumable-media', 'google-auth', 'protobuf']
tests ['smdebug-rulesconfig', 'testcontainers']
ruamel ['ruamel.yaml.clib', 'ruamel.yaml']
```

PR #525, opened 2022-12-12
Find correct package names

### LINKED ISSUE #362: Add config option to add alternative python packages
Sometimes python package name is not the same as it's module name (eg `sklearn` vs `scikit-learn`). For that we have `MODULE_PACKAGE_MAPPING` in `mlem.core.requirements` that you can update from code. But it would be nice to be able to do it from config like
`config.yaml`
```
core:
  package_mapping:
    box: python-box
```

aguschin:
close https://github.com/iterative/mlem/issues/362

some things still aren't fixed, but IMO it's low priority now #524

### DIFF
--- a/mlem/core/requirements.py
+++ b/mlem/core/requirements.py
@@ -25,7 +25,7 @@ from typing import (
     Union,
 )
 
-from pydantic import BaseModel
+from pydantic import BaseModel, root_validator
 
 from mlem.core.base import MlemABC
 
@@ -85,16 +85,17 @@ class InstallableRequirement(PythonRequirement):
     package_name: Optional[str] = None
     """Pip package name for this module, if it is different from module name"""
 
-    def __init__(self, **kwargs):
-        from mlem.utils.module import get_package_name
+    @root_validator
+    def set_package_name(cls, values):  # pylint: disable=no-self-argument
+        package_name = values.get("package_name")
+        if not package_name:
+            from mlem.utils.module import get_package_name
 
-        # Q: is this ok or I should specify args explicitly?
-        super().__init__(**kwargs)
-        if (
-            not self.package_name
-            and get_package_name(self.module) != self.module
-        ):
-            self.package_name = get_package_name(self.module)
+            module = values.get("module")
+            package_name = get_package_name(module)
+            if package_name != module:
+                values["package_name"] = package_name
+        return values
 
     @property
     def package(self):
--- a/mlem/core/requirements.py
+++ b/mlem/core/requirements.py
@@ -87,16 +87,20 @@ class InstallableRequirement(PythonRequirement):
 
     def __init__(self, **kwargs):
         from mlem.utils.module import get_package_name
+
         # Q: is this ok or I should specify args explicitly?
         super().__init__(**kwargs)
-        if not self.package_name and get_package_name(self.module) != self.module:
+        if (
+            not self.package_name
+            and get_package_name(self.module) != self.module
+        ):
             self.package_name = get_package_name(self.module)
 
     @property
     def package(self):
         """
         Pip package name
-        """        
+        """
         return self.package_name or self.module
 
     def get_repr(self):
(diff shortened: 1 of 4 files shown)


mlem/utils/module.py
```diff
@@ -266,8 +266,10 @@ def get_package_name(mod: ModuleType) -> str:
     """
     if mod is None:
         raise ValueError("mod must not be None")
-    name = mod.__name__
-    return MODULE_PACKAGE_MAPPING.get(name, name)
+    if isinstance(mod, str):
+        mod = importing.import_module(mod)
+    # QUESTION: should we issue a warning if this list have > 1 item?
```
aguschin, line 271: My take is to print a warning for now, but need to investigate in future to resolve
https://github.com/iterative/mlem/issues/524#issuecomment-1346420956

mlem/core/requirements.py
```diff
@@ -92,14 +85,23 @@ class InstallableRequirement(PythonRequirement):
     package_name: Optional[str] = None
     """Pip package name for this module, if it is different from module name"""
 
+    def __init__(self, **kwargs):
+        from mlem.utils.module import get_package_name
+
+        # Q: is this ok or I should specify args explicitly?
```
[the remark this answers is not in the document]
mike0sv: thats ok,maybe make it like original signature `**data: Any`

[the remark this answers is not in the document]
mike0sv: but actually, use `@validator` classmethod  for this, not constructor

mlem/ext.py
```diff
@@ -59,9 +60,8 @@ def __hash__(self):
 
     @property
     def reqs_packages(self):
-        from mlem.core.requirements import MODULE_PACKAGE_MAPPING
-
-        return [MODULE_PACKAGE_MAPPING.get(r, r) for r in self.reqs]
+        # Q: now this requires for packages to be installed. Can it break for some reason?
```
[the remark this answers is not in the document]
mike0sv: yes, because it is used exactly when you don't have required packages to print msg like `to install them run `pip install ...`

mlem/utils/module.py
```diff
@@ -266,8 +266,10 @@ def get_package_name(mod: ModuleType) -> str:
     """
     if mod is None:
         raise ValueError("mod must not be None")
-    name = mod.__name__
-    return MODULE_PACKAGE_MAPPING.get(name, name)
+    if isinstance(mod, str):
+        mod = importing.import_module(mod)
```
mike0sv, line 270: This will force to import module every time. Can you switch to `if not str: mod = mod.__name__`?

mlem/utils/module.py
mike0sv: I think you should print a warning of course, but if you print it you also should provide a way to resolve it. Eg manually configuring the mapping

mlem/core/requirements.py
```diff
@@ -92,14 +85,24 @@ class InstallableRequirement(PythonRequirement):
     package_name: Optional[str] = None
     """Pip package name for this module, if it is different from module name"""
 
+    @root_validator
```
mike0sv, line 88: why not `@validator("pachage_name")`?

mlem/ext.py
```diff
@@ -59,9 +59,12 @@ def __hash__(self):
 
     @property
     def reqs_packages(self):
-        from mlem.core.requirements import MODULE_PACKAGE_MAPPING
-
-        return [MODULE_PACKAGE_MAPPING.get(r, r) for r in self.reqs]
+        # since this must work without these packages installed,
+        # we use mapping for cases when module_name != package_name
+        module_package = {
```
mike0sv, line 64: maybe just return the mapping constant for this one?

mlem/core/requirements.py
aguschin: because you need `module` as well - and it won't be accessible that way

mlem/ext.py
aguschin: Sure

aguschin: but, on other hand, this is a single place where it's used. I can make a constant, but what's a point?

mike0sv (approved):

PR #527, opened 2022-12-13
Add fastapi to `pip install mlem[streamlit]`

aguschin:
@mike0sv, IIUC it should work like this. Could you please check?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -70,6 +70,7 @@ extras = {
     "xgboost": ["xgboost"],
     "lightgbm": ["lightgbm"],
     "fastapi": ["uvicorn", "fastapi"],
+    "streamlit": ["uvicorn", "fastapi", "streamlit", "streamlit_pydantic"],
     "sagemaker": ["boto3", "sagemaker"],
     "torch": ["torch"],
     "tensorflow": ["tensorflow"],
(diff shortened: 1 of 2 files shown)


mike0sv:
I initially did it like this, but decided to make those optional for streamlit. If you check streamlit server get_requirements, fastapi reqs are added if you dont have external backend address.

We can make them required though, but you also need to fix `ext.py` I guess

aguschin:
Got your idea. IMO we still better install `fastapi` with `pip install mlem[streamlit]`, since otherwise >90% of users will encounter errors and have to do it manually. Even if it's not always necessary requirement.

setup.py
```diff
@@ -70,6 +70,7 @@
     "xgboost": ["xgboost"],
     "lightgbm": ["lightgbm"],
     "fastapi": ["uvicorn", "fastapi"],
+    "streamlit": ["uvicorn", "fastapi", "streamlit", "streamlit_pydantic"],
     "sagemaker": ["boto3", "sagemaker"],
```
aguschin, line 74: maybe:
```suggestion
    "sagemaker": ["docker", "boto3", "sagemaker"],
```
? @mike0sv 
or you had some idea behind this as well?

mike0sv: probably not :)

mike0sv (approved):

PR #534, opened 2022-12-20
importlib_metadata dependency

mike0sv:


aguschin:
great! I see one issue in failing test - it doesn't check packages like `package>1` or `package<0.0.1`, only packages like `package` or `package==0.0.1`. I'll fix that test a bit later and will merge this.

PR #553, opened 2022-12-29
Add huggingface transformers support

mike0sv:


aguschin:
Hit it again - saving model doesn't save `torch` if I used transformers - same as in https://medium.com/better-programming/i-trained-a-model-to-tell-if-you-were-naughty-this-year-11a36ca6d472

I forgot, do we fix it in this PR? I hope we do :)

PR #558, opened 2022-12-30
GitHub actions update

dacbd:
uses the new output file.
https://docs.github.com/en/actions/using-jobs/defining-outputs-for-jobs


aguschin (approved):
thanks!

PR #561, opened 2022-12-30
Update GH workflow

aguschin:
1. This needs to be checked by creating a dummy PR that updates this branch. To run the workflow here.
2. This breaks CI - see https://github.com/iterative/mlem/actions/runs/3808281067/jobs/6478727408. Need to be fixed before merge

### DIFF
--- a/.github/workflows/check-test-release.yml
+++ b/.github/workflows/check-test-release.yml
@@ -75,12 +75,12 @@ jobs:
     - name: get pip cache dir
       id: pip-cache-dir
       run: |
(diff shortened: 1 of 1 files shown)


aguschin:
Eh, the problem is for Windows - looks like `echo "PIP_CACHE_DIR=$(pip cache dir)" >> $GITHUB_ENV` doesn't add an env var.

skshetry:
@aguschin, `setup-python` has in-built caching now, you can get rid of most of the code here:

https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#caching-packages.

PR #582, opened 2023-01-19
Fixing small typo in the CLI

### LINKED ISSUE #579: Typo (?): standartize
`mlem serve` has the option `--standartize` which I suppose should be `--standardize`.   Seems like not too big of an issue and fixing would be a breaking change... but thought I would point that out!  Thanks for your work on this.

keithing:
This fixes a typo in the `--standartize` argument of `mlem serve`.  A few details:

- The way I've implemented it, `mlem serve` now accepts both `--standardize/--no-standardize` and `--standartize/--no-standartize` to avoid breaking changes.  Fortunately this is pretty simple to do with `typer`.
- I noticed that before this change, the variable `standartize` was being passed to `Server` via `**kwargs`, while `Server` [was expecting the parameter `standardize`](https://github.com/iterative/mlem/blob/c7f91613b7a71a09dcbfb79c6254346f181d4a32/mlem/runtime/server.py#L174). So passing `--no-standardize` actually had no effect on the server.  Not sure how important that is 🤷 
- I'm happy to add tests, but the logic seems a little tangled and tricky to test and I could use a little help if you think testing is worth it.

Closes #579

### DIFF
--- a/mlem/cli/serve.py
+++ b/mlem/cli/serve.py
@@ -33,12 +33,13 @@ serve = Typer(
 )
 app.add_typer(serve)
 
-option_standartize = Option(
+# Include "--standartize" for historical compatability
+option_standardize = Option(
     True,
+    "--standardize/--no-standardize",
     "--standartize/--no-standartize",
     help="Whether to conform model interface to standard ('predict' method with single arg 'data')",
 )
-
 
 @mlem_group_callback(serve, required=["model", "load"])
 def serve_load(
@@ -46,7 +47,7 @@ def serve_load(
     project: Optional[str] = option_project,
     rev: Optional[str] = option_rev,
     load: Optional[str] = option_load("server"),
-    standartize: bool = option_standartize,
+    standardize: bool = option_standardize,
 ):
     from mlem.api.commands import serve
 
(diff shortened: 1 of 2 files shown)


mlem/cli/serve.py
```diff
@@ -33,20 +33,21 @@
 )
 app.add_typer(serve)
 
-option_standartize = Option(
+# Include "--standartize" for historical compatability
+option_standardize = Option(
     True,
+    "--standardize/--no-standardize",
     "--standartize/--no-standartize",
```
aguschin, line 40:
```suggestion
    "--standartize/--no-standartize",  # deprecate in future versions
```

mlem/cli/serve.py
```diff
@@ -33,20 +33,21 @@
...
     help="Whether to conform model interface to standard ('predict' method with single arg 'data')",
 )
```
aguschin, line 43:
```suggestion

```

aguschin (approved):
Thanks @keithing! 🚀 Will merge when tests pass.

keithing:
Perfect. Looks like you restarted the test, which may have been an intermittent error.  If it's actually an error produced by my code I'll keep an eye out and fix as needed.

mike0sv (approved):

PR #600, opened 2023-01-27
546 boolean options flags

### LINKED ISSUE #546: Error with `--current_env` in `mlem build venv`
The docs indicate the option is boolean:
`--current_env: Whether to install in the current virtual env, must be active [default: False]`
but:
```
$ mlem build venv --model lyrics2emoji --current_env
╭─ Error ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ Option '--current_env' requires an argument.                                                                                                         │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
$ mlem build venv --model lyrics2emoji --current_env True
⏳️ Loading model from lyrics2emoji.mlem
💼 Detected the virtual env /Users/aguschin/Git/iterative/mlem/emoji/.venv
💼 Installing the required packages...
......
```

ykasimov:
Fixes #546

### DIFF
--- a/mlem/cli/utils.py
+++ b/mlem/cli/utils.py
@@ -508,6 +508,7 @@ def _option_from_field(
         type_ = anything(type_)
     elif field.allow_none:
         type_ = optional(type_)
+    # breakpoint()
     option = SetViaFileTyperOption(
         param_decls=[f"--{path}", path.replace(".", "_")],
         type=type_ if not force_not_set else anything(type_),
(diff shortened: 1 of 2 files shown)


aguschin:
Hi @ykasimov! Approved PR run, and added you once again to the reviewers. You'll get notifications for PRs in this repo, but you are able now to approve yours :)

aguschin:
@mike0sv is this ready to be merged?

mike0sv:
@ykasimov you still need to address the second part of my comment in the original issue

Need to be careful though because we also might need to alter option name. Eg if is_on has default value False, it makes sense that --is_on will make it True. But if is_on default value is True, we should have something like --not_is_on as an option name

Otherwise you either will not be able to change default value via CLI or it will be name for a flag. Also need some tests for this

PR #601, opened 2023-02-02
Add no-deploy option for Flyio app creation

mike0sv:


mlem/contrib/flyio/meta.py
```diff
@@ -101,6 +101,7 @@ def _create_app(self, state: FlyioAppState):
                 "region": self.region
                 or self.get_env().region
                 or project_config("", section=FlyioConfig).region,
+                "no-deploy": True,
```
aguschin, line 104: What this "no-deploy" option does?

mike0sv: By default this command will try to actually run something. But at this stage we only want to create app and not run it (we don't have docker dir yet)

aguschin (approved):

PR #624, opened 2023-02-28
Fix `stata` saving

### LINKED ISSUE #618: reset_index inside write method of PandasFormat class crashes the reader for stata format
reset_index inside write method of PandasFormat class results in adding an index column with an empty label. This crashes the stata write method in pandas library which checks for the invalid column names. I am not sure why one needs to have reset_index method at all. It adds the index as a new column. Removing the function call to reset_index solves the issue

aguschin:
close #618

### DIFF
--- a/mlem/contrib/pandas.py
+++ b/mlem/contrib/pandas.py
@@ -305,6 +305,11 @@ def has_index(df: pd.DataFrame):
     return not isinstance(df.index, pd.RangeIndex)
 
 
+def has_named_index(df: pd.DataFrame):
+    """Returns true if all index columns are named"""
+    return df.index.name or all(df.index.names)
+
+
 def _reset_index(df: pd.DataFrame):
     """Transforms indexes to columns"""
     index_name = df.index.name or ""  # save it for future renaming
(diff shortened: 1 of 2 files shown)


mlem/contrib/pandas.py
```diff
@@ -459,7 +459,10 @@ def write(
             write_kwargs.update(self.write_args)
         write_kwargs.update(kwargs)
 
-        if has_index(df):
+        if (
+            has_index(df)
+            and PANDAS_FORMATS["stata"].write_func != self.write_func
```
aguschin, line 464: This problem occurs only with stata format, and only in the case described in original issue (i.e. when you import stata format file and want to re-write initial content - cause otherwise the dataframe will be saved in `csv` and no problem occurs at write). In this case there could be no real index (since stata format doesn't support saving index in some special way - it's going to be saved as a column). Then at import MLEM cannot have an index in this dataframe, which allows us to skip resetting it.

I can't see any solution better than creating a workaround like: if you have empty index, rename column to something like `"__index__"`, and save the instruction to rename it on load back to `""`. Which is something I'd like to avoid now since we don't write to `stata` format anyway for now - only to csv (which means this problem just won't occur except for special case I mentioned in the first paragraph).

omesser: Like the original reporter in the issue, I don't completely know why we're resetting index here. Any idea @aguschin ?

aguschin: You may use index in the model itself. E.g. the index may have some information used by the model directly, like customer id, timestamp, etc. So, to make things more reproducible and precise, we decided to keep the index. We argued about this with @mike0sv a while ago, so that's a decision made early on. We can debate again whether this is a bad practice and/or maybe change the default behavior: not store index by default.

omesser: Makes sense. This is the sort of things usually code comments are used for - explain why we're doing non-obvious operations (can link to gh isues/discussions, etc)

omesser (approved):
Suggest to add comments - why reset index, and now adding a comment for the condition linking to the issue

PR #627, opened 2023-03-03
contrib: fix importing pandas datasets in excel format

### LINKED ISSUE #617: importing excel files incorrectly sets the format to csv and writes data in csv
if one tries to import an excel file using mlem.api.import_object() function and sets type_ as "pandas[excel]" MLEM sets the format  incorrectly to csv and also if the copy_data is set to True it will write the data in csv format instead of excel.
I have tracked down the issue and the problem lies in two classes "DataFrameType" and "SeriesType" in contrib.pandas package
in _get_writer()_ method of these two classes it compares the extension of the file which is in this case either "xlsx" or "xls" with the list of extensions in _PANDAS_FORMATS_ and _PANDAS_SERIES_FORMATS_ and since it does not match with "excel" it defaults to the csv which is the default format.
One solution would be to add an extra elif such as 
```python
if ext in PANDAS_SERIES_FORMATS:
    fmt = ext
elif ext in ["xls", "xlsx"]:
    fmt = "excel"
```

aliahari:
Fixes #617

### DIFF
--- a/mlem/contrib/pandas.py
+++ b/mlem/contrib/pandas.py
@@ -270,6 +270,8 @@ class SeriesType(WithDefaultSerializer, _PandasDataType):
             ext = ext.lstrip(".")
             if ext in PANDAS_SERIES_FORMATS:
                 fmt = ext
+            elif ext in ["xls", "xlsx"]:
+                fmt = "excel"
         return PandasSeriesWriter(format=fmt)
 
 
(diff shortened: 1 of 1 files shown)


aguschin:
Thanks! Do you mind adding few tests for this in https://github.com/iterative/mlem/blob/main/tests/contrib/test_pandas.py?

aguschin:
I think we can merge this. Looks like it should be working. 
Thanks for contribution @aliahari!

aguschin (approved):

PR #633, opened 2023-03-09
Add new DVC extra

mike0sv:


aguschin (approved):

setup.py
```diff
@@ -110,6 +110,7 @@
     "webdav",
     "webhdfs",
     "webhdfs_kerberos",
+    "lint",
```
skshetry, line 113: Why is this needed? We usually move these things around without a guarantee that they will always work or be available (other than remote related extras).

aguschin: This is to enable `pip install mlem[dvc-s3]` and ones like that. We have a test that checks we didn't miss anything, that's why we need to add keep it here.

aguschin: btw, I think we should've added it here instead https://github.com/iterative/mlem/blob/d2100acaa8b092bc416cfc65ab0583a90b2e776d/tests/test_setup.py#L18

aguschin: doing that in https://github.com/iterative/mlem/pull/634

Issue #463, opened 2022-10-31
`mlem deploy run` should be able to re-deploy the model regardless of state


aguschin:
I stumble upon this each time I re-generate http://github.com/iterative/example-mlem-get-started

To be able to re-deploy, we use this:
```shell
if heroku apps:info example-mlem-get-started-app; then
  heroku apps:destroy example-mlem-get-started-app --confirm example-mlem-get-started-app
  heroku container:login
fi
```
Which is fine for some cases, but I would like to have a `--force` or `--ignore-state` flag, to just re-deploy without falling back to Heroku CLI first.

Besides, when I remove the deployment outside of MLEM, there is a time frame when app is not available. I suppose, Heroku should handle this gracefully, but there is no way since I need to remove the app.

Finally, app settings are removed (e.g. I need to change Dyno to Hobby now each time manually). Not convenient :)

alex000kim:
After logging in, most users will try rerunning the same command, so instead of throwing an error, we should suggest users either re-use the deployment configuration (`app.mlem`) or overwrite it.
```
$ mlem deployment run heroku app.mlem --model models/clf-model  --app_name bank-churn-app
:floppy_disk: Saving deployment to app.mlem
:hourglass_flowing_sand:️ Loading model from models/clf-model.mlem
🏛 Creating Heroku App bank-churn-app
 ›   Error: Invalid credentials provided.
 ›
 ›   Error ID: unauthorized
:x: Invalid credentials provided.

$ heroku login
 ›   Warning: heroku update available from 7.63.4 to 7.66.4.
heroku: Press any key to open up the browser to login or q to exit: 
Opening browser to https://cli-auth.heroku.com/auth/cli/browser/...
Logging in... done
Logged in as ...

$ mlem deployment run heroku app.mlem --model models/clf-model  --app_name bank-churn-app
:hourglass_flowing_sand:️ Loading deployment from app.mlem
:x: Deployment meta already exists at path=.../demo-bank-customer-churn/app.mlem' project=None rev=None 
uri='.../demo-bank-customer-churn/app.mlem' project_uri=None 
fs=<fsspec.implementations.local.LocalFileSystem object at 0x101ff9c40>. Please use `mlem deployment run --load <path> ...`
```

mike0sv:
Probably we can also add `--force` flag (or `--overwrite`?)

alex000kim:
That's def an option. 
[...]

PR #644, opened 2023-03-16
Allow to force deploy if deploy parameters change

### LINKED ISSUE #549: Changing parameters of deployment isn't allowed
If you created a deployment, but then trying to change it's params, MLEM won't allow you until you remove `.state` file:

```shell
$ mlem deploy run docker_container dc.mlem -m lyrics2emoji --image_name lyrics2emoji
deployed
$ mlem deploy remove dc
removed
$ rm dc.mlem
$ mlem -tb deploy run docker_container dc.mlem -m lyrics2emoji --image_name lyrics2emoji --ports.0 8080:8080
...
mlem.core.errors.DeploymentError: Deployment parameters changed, this is not supported yet. Please re-create deployment with new parameters
```

Until you realize you need to remove `dc.mlem.state` file you'll be getting this problems. You can need to use different  declaration files until you realize thet and delete them.

TBH I would expect it can be done even without `mlem deploy remove dc` and `rm dc.mlem`.

Can we enable this? So that running the deploy again will rewrite `dc.mlem` and re-deploy the model again?

```shell
$ mlem deploy run docker_container dc.mlem -m lyrics2emoji --image_name lyrics2emoji
deployed
$ mlem deploy run docker_container dc.mlem -m lyrics2emoji --image_name lyrics2emoji --ports.0 8080:8080
dc.mlem updated and service re-deployed
```

I understand this can be justified by usage safety - e.g. no way to redeploy a model with different params unless you know what you doing. But for novice users this can be a problem - they can skip MLEM because it's not intuitive here, I need to somehow figure out how do I redeploy my model, etc. Can be confusing.

mike0sv:
Adds `--force` flag to deploy cli command. If deployment parameters changed and this flag set, mlem will remove old deployment and create new one.
Also, now there will be no error if parameters changed, but old app does not have state

Limitations:
* for now only available from cli since adding this to API required a bit of refactoring
* deployment will always be re-created even if platform allows to just update old deployment

related #463
closes #549

### DIFF
--- a/mlem/cli/deployment.py
+++ b/mlem/cli/deployment.py
@@ -44,7 +44,7 @@ from mlem.core.objects import (
     MlemModel,
 )
 from mlem.telemetry import pass_telemetry_params
-from mlem.ui import echo, no_echo, set_echo
+from mlem.ui import EMOJI_STOP, echo, no_echo, set_echo
 
 logger = logging.getLogger(__name__)
 
@@ -116,6 +116,9 @@ def create_deploy_run_command(type_name):
         model_rev: Optional[str] = option_model_rev,
         project: Optional[str] = option_project,
         file_conf: List[str] = option_file_conf("deployment"),
+        force: bool = Option(
+            False, is_flag=True, help="Force re-create if parameters changed"
+        ),
         **__kwargs__,
     ):
         from mlem.api.commands import deploy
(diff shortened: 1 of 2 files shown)


aguschin (approved):
Thanks @mike0sv! I assume this is good to be merged, any subsequent work (if any) will be in next PRs?

omesser:
Can this be merged? 👀

Issue #661, opened 2023-04-26
flyctl deployment failure: org slug must be specified when not running interactively


aguschin:
Deployment to fly fails when you're in multiple orgs (happen always - this is blocker, in fact):
1. As example below will show, passing org slug doesn't work. We need to fix it
2. There is no explanation in MLEM docs about this. We can't hide complexity of deploy without explaining how to fix such errors. Need to:
    - update docstring for the option suggesting to run `flyctl orgs list` to get the right slug
    - explain this in docs (user guide)

```
$ mlem deploy run flyio app.mlem -m model --app_name test-app-mlem
⏳️ Loading deployment from app.mlem
💾 Saving deployment to app.mlem
⏳️ Loading model from model.mlem
Error: org slug must be specified when not running interactively

❌ Command '['flyctl', 'launch', '--auto-confirm', '--region', 'lax', '--no-deploy', '--name', 'test-app-mlem']' returned non-zero exit status 1.
```
ok, let's find out slug:
```
$ flyctl orgs list
Name                 Slug                 Type
----                 ----                 ----
Alexander Guschin    personal             PERSONAL
...
```
ok, let's use it:
```
$ mlem deploy run flyio app.mlem -m model --app_name test-app-mlem --env.org personal
⏳️ Loading deployment from app.mlem
💾 Saving deployment to app.mlem
⏳️ Loading model from model.mlem
Error: org slug must be specified when not running interactively

❌ Command '['flyctl', 'launch', '--auto-confirm', '--region', 'lax', '--no-deploy', '--name', 'test-app-mlem']' returned non-zero exit status 1.

$ mlem deploy run flyio app.mlem -m model --app_name test-app-mlem --env.org personal --org personal
⏳️ Loading deployment from app.mlem
💾 Saving deployment to app.mlem
⏳️ Loading model from model.mlem
Error: org slug must be specified when not running interactively

❌ Command '['flyctl', 'launch', '--auto-confirm', '--region', 'lax', '--no-deploy', '--name', 'test-app-mlem']' returned non-zero exit status 1.
```

For the reference:
```
$ mlem deploy run flyio --help

 Usage: mlem deploy run flyio [options] path

 fly.io deployment
 Documentation: <https://mlem.ai/doc/command-reference/deployment>

╭─ Arguments ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ *    path      TEXT  Where to save the object (.mlem file) [default: None] [required]                                                                │
[...]

Issue #657, opened 2023-04-26
`mlem deploy flyio` fails due to wrong `fly.toml` generated


aguschin:
Under the hood [MLEM calls](https://github.com/iterative/mlem/blob/main/mlem/contrib/flyio/meta.py#L113) `flyctl launch` to generate `fly.toml` config for deployment.

It happens, that sometimes `flyctl launch` generates very short `fly.toml` instead of the expected one. That file doesn't have an instruction for fly to work with `8080` port (or whichever one we specify) and after deployment service doesn't respond to requests.

example of "short"/"wrong" fly.toml:
```toml
app = "art-expert-retry2"
kill_signal = "SIGINT"
kill_timeout = 5
primary_region = "lax"
processes = []
```
example of "long"/"correct" one:
```toml
app = "art-expert-retry2"
kill_signal = "SIGINT"
kill_timeout = 5
primary_region = "lax"
processes = []

[env]

[experimental]
  auto_rollback = true

[[services]]
  http_checks = []
  internal_port = 8080
  processes = ["app"]
  protocol = "tcp"
  script_checks = []
  [services.concurrency]
    hard_limit = 25
    soft_limit = 20
    type = "connections"

  [[services.ports]]
    force_https = true
    handlers = ["http"]
    port = 80

  [[services.ports]]
    handlers = ["tls", "http"]
    port = 443

  [[services.tcp_checks]]
    grace_period = "1s"
    interval = "15s"
    restart_limit = 0
    timeout = "2s"
```

Quick search didn't give me any hint whether we simply don't specify some argument for `flyctl launch` or whether this is a bug in fly we need to wait a fix for.

PR #662, opened 2023-04-26
Fix flyio org option not working

aminalaee:
Subset of https://github.com/iterative/mlem/issues/661

### DIFF
--- a/mlem/contrib/flyio/meta.py
+++ b/mlem/contrib/flyio/meta.py
@@ -44,7 +44,7 @@ class FlyioConfig(MlemConfigBase):
 
 class FlyioSettings(BaseModel):
     org: Optional[str] = None
-    """Organization name"""
+    """Organization slug (run `flyctl orgs list` to find out the right one)"""
     region: Optional[str] = None
     """Region name"""
 
--- a/mlem/contrib/flyio/meta.py
+++ b/mlem/contrib/flyio/meta.py
@@ -110,6 +110,8 @@ class FlyioApp(MlemDeployment, FlyioSettings):
                 args["generate-name"] = True
             if self.get_env().access_token:
                 args["access-token"] = self.get_env().access_token
+            if self.org:
+                args["org"] = self.org
             run_flyctl("launch", workdir=tempdir, kwargs=args)
             state.fly_toml = read_fly_toml(tempdir)
             port = getattr(self.server, "port", None) or getattr(
--- a/.github/workflows/check-test-release.yml
+++ b/.github/workflows/check-test-release.yml
@@ -99,6 +99,11 @@ jobs:
       env:
         HEROKU_API_KEY: ${{ secrets.HEROKU_API_KEY }}
         HEROKU_TEAM: iterative-sandbox
+    - name: Run Flyctl tests
+      if: matrix.os == 'ubuntu-latest' && matrix.python == '3.9'
+      run: |
+        curl -L https://fly.io/install.sh | sh
+        pytest -k 'flyctl'
     - name: Start minikube
       if: matrix.os == 'ubuntu-latest' && matrix.python == '3.9'
       uses: medyagh/setup-minikube@master
--- a/tests/contrib/test_flyio.py
+++ b/tests/contrib/test_flyio.py
@@ -1,0 +1,34 @@
+from pathlib import Path
+from unittest.mock import ANY, patch
+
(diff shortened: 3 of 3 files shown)


mlem/contrib/flyio/meta.py
```diff
@@ -110,6 +110,8 @@ def _create_app(self, state: FlyioAppState):
                 args["generate-name"] = True
             if self.get_env().access_token:
                 args["access-token"] = self.get_env().access_token
+            if self.org:
+                args["org"] = self.org
```
aminalaee, line 114: I couldn't find any tests,

aguschin: Good point, looks like we didn't add them 👀

aguschin: Is this something you're interested in? As a first step I think we should add a test for `_create_app`, asserting that resulting `fly.toml` looks right. This test can also help us pin down the root cause of https://github.com/iterative/mlem/issues/657

The second step would be to add a real e2e deploy to fly. I can create a test account there and set up GHA workflow to make it work.

aguschin (commented):
Edit docstring

mlem/contrib/flyio/meta.py
aminalaee: Yeah sounds good. Do you want to add the tests in a separate PR?

mlem/contrib/flyio/meta.py
```diff
@@ -44,7 +44,7 @@ class Config:
 
 class FlyioSettings(BaseModel):
     org: Optional[str] = None
-    """Organization name"""
+    """Organization slug (run `flyctl orgs list` to find out the right one)"""
```
aguschin, line 47: This will update help messages we see in `mlem deploy run flyio --help`

mlem/contrib/flyio/meta.py
aguschin: No, this PR is the right place :)

aminalaee: @aguschin Any idea how this can be tested?
I'm doing
```py
def test_create_app():
    flyio_app = FlyioApp()
    state = flyio_app.get_state()
    flyio_app._create_app(state)
```

But the `get_state()` fails because of `mlem.core.errors.MlemObjectNotSavedError: Not saved object has no location`.
even if I create the app with a default state, while updating the state the issue happens again. I think I'm missing a step before the the call to create_app.

aguschin: Hi @aminalaee! Sure, let me help you out 🙌🏻 
It needs to be dumped first:
```py
from mlem.contrib.flyio.meta import FlyioApp


def test_create_app(tmpdir):
    flyio_app = FlyioApp()
    flyio_app.dump(str(tmpdir))
    state = flyio_app.get_state()
    flyio_app._create_app(state)
```

If you run this test, it will fail still, because creating flyio app requires Dockerfile present (`...Could not find a Dockerfile, nor detect a runtime or framework from source code. Continuing with a blank app.\x1b`).

aguschin: You either need to bypass this, or maybe call `mlem.api.deploy` directly on `flyio_app` while mocking some actual deployment mechanisms. Then, until the temp dir with `fly.toml` and stuff is removed, you can read it in the test to make some checks. Hope that makes it clearer! If confused, please ping me, I'll help with that.

aguschin:
Hi @aminalaee! Overall, LGTM!

Two comments:
1. I see `FAILED tests/contrib/test_flyio.py::test_create_app - FileNotFoundError: [Errno 2] No such file or directory: 'flyctl'` - we need to edit workflow to install `flyctl` and maybe even login with it - we need to see if running the test is possible without login
2. Would be great to add a test that checks `fly.toml` content generated by flyctl (can be done in another PR though).

aminalaee:
I will update the actions to install the flyctl cli.
But as for the login, in this test I have mocked the calls so that we don’t need credentials as we don’t call their API.

but this limits us for number 2 which adds an integration test. We can add another one.

aminalaee:
Hey @aguschin 
Sorry I'd missed the email and it was treated as spam (weird). I will check that now.

I was thinking to fix the pipeline in this PR with the current mocked tests and setup the integration tests in a separate PR?

aguschin (approved):

PR #669, opened 2023-05-01
Add more test decorators

aminalaee:
Not sure if this is accepted but some tests fail locally when credentials/env variables are not set.
I see that there are already decorators for this purpose, this PR extends those decorators.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -5,7 +5,7 @@ from mlem.constants import MLEM_CONFIG_FILE_NAME
 from mlem.contrib.fastapi import FastAPIServer
 from mlem.core.artifacts import FSSpecStorage, LocalStorage
 from mlem.core.meta_io import get_fs
-from tests.conftest import long
+from tests.conftest import long, need_aws_auth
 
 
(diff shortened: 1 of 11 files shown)


tests/test_config.py
```diff
@@ -21,6 +21,7 @@ def test_loading_empty(set_mlem_project_root):
     assert isinstance(config.storage, LocalStorage)
 
 
+@need_aws_auth
 @long
 def test_loading_remote(s3_tmp_path, s3_storage_fs):
     project = s3_tmp_path("remote_conf")
```
aguschin, line 27: Looks good, thank you @aminalaee. This should make it easier contributing 🙏🏻  The only question - I thought some tests with aws will pass since to **read** test bucket, you don't need any special permissions. Is this wrong?

aminalaee: Not sure the case, but `boto3.Session().get_credentials()` will return None if the AWS credentials `~/.aws/credentials` doesn't exist on my local. I guess this is even before making the calls to AWS services.

aguschin (approved):
LGTM, merging! Thanks a lot @aminalaee !

PR #671, opened 2023-05-10
Checking flyio org PR didn't break anything

aguschin:
@aminalaee, when you change GH workflow in a PR, GH runs the workflow from the `base` branch. So, we couldn't test your changes. This is not recommended OFC, but I didn't found a good way to check them on your PR (usually I create a dummy branch that changes README or something, and set `base` for it to be feature branch that changes workflow). So I merged, since your PR LGTM, and in this PR we're actually checking it didn't break anything 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,5 @@
 ![image](https://user-images.githubusercontent.com/6797716/165590476-994d4d93-8e98-4afb-b5f8-6f42b9d56efc.png)
 
-
 [![Check, test and release](https://github.com/iterative/mlem/actions/workflows/check-test-release.yml/badge.svg)](https://github.com/iterative/mlem/actions/workflows/check-test-release.yml)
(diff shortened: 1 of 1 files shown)


aminalaee:
Hmm...strange. Is it failing now because of flyctl?

aguschin:
Nope, `authorize` job failed yesterday - not sure what was the cause. Re-run the job, it passed, now let's see how other jobs finish.

aminalaee:
Now it failed, I think the reasons were:
- I should rename the test to `test_flyio_create_app`
- And in the GH action it should be `pytest -k flyio` not flyctl 🤦 
- And not sure if flyctl should be added to the $PATH or it will work out of the box after install

aguschin:
Ok, no problem @aminalaee. Let me give you rights to write to this repo, so we can test it properly now, without merging things in `main` first. I sent you an invite. Can you create a branch with fix **in this repo** (not using your fork), and make a PR to `main`? Then I'll create a dummy branch to be "merged" in your branch, and there we'll see how CI finishes.

aminalaee:
Ok sure, @aguschin Thanks

PR #672, opened 2023-05-11
Fix Flyio GH action

aminalaee:
Fix Flyio Github action

### DIFF
--- a/.github/workflows/check-test-release.yml
+++ b/.github/workflows/check-test-release.yml
@@ -99,17 +99,18 @@ jobs:
       env:
         HEROKU_API_KEY: ${{ secrets.HEROKU_API_KEY }}
         HEROKU_TEAM: iterative-sandbox
-    - name: Run Flyctl tests
+    - name: Setup Flyio
+      uses: superfly/flyctl-actions/setup-flyctl@master
       if: matrix.os == 'ubuntu-latest' && matrix.python == '3.9'
-      run: |
-        curl -L https://fly.io/install.sh | sh
-        pytest -k 'flyctl'
+    - name: Run Flyio tests
+      if: matrix.os == 'ubuntu-latest' && matrix.python == '3.9'
+      run: pytest -k 'flyio'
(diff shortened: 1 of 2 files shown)


.github/workflows/check-test-release.yml
```diff
@@ -99,17 +99,18 @@ jobs:
...
+    - name: Run Flyio tests
+      if: matrix.os == 'ubuntu-latest' && matrix.python == '3.9'
+      run: pytest -k 'flyio'
     - name: Start minikube
       if: matrix.os == 'ubuntu-latest' && matrix.python == '3.9'
       uses: medyagh/setup-minikube@master
     - name: Run tests
       timeout-minutes: 40
-      run: pytest -k 'not heroku'
+      run: pytest -k 'not heroku and not flyctl'
```
aguschin, line 113: > https://github.com/iterative/mlem/blob/main/tests/contrib/test_heroku.py#L37-L39

I assume:
```suggestion
      run: pytest -k 'not heroku and not flyio'
```

aguschin, line 113: > FAILED tests/contrib/test_flyio.py::test_flyio_create_app - FileNotFoundError: [Errno 2] No such file or directory: 'flyctl'

I assume:
```suggestion
      run: pytest -k 'not heroku and not flyio'
```

aminalaee:
You are right about `pytest -k 'not heroku and not flyctl'` but I think the dummy PR passed the pipeline. This PR is still using the old action, right?

aguschin:
> I think the dummy PR passed the pipeline

Yes, looks like that!

> This PR is still using the old action, right?

Yes.

aguschin:
Yes, CI finished. There was some error with catboost, I've seen it in earlier PRs. We'll need to fix it, if it doesn't take care of itself, in order to release next version of MLEM. Merging! Thanks for the quick fix @aminalaee !

aguschin (approved):
LGTM 👍🏻

PR #681, opened 2023-05-30
Expose GTO version in FastAPI's interface

### LINKED ISSUE #665: Expose GTO's model version in FastAPI's `interface.json`
Follow-up for #664. Sometimes it's desired to store predictions along with the specific model version that returned those predictions. There are at least two ways to support that in MLEM:
1. Return model version in prediction - then what's returned will be a json like `{"prediction": [0.4, 0.6], "version": 0.1.3}`. I've seen some generic ML frameworks doing this IIRC.
2. Return it in `interface.json` - we already have MLEM version there, so adding model version looks logical
image

Regarding how we get this info into the service. Again, there are two approaches:
1. Add it at `mlem.api.save`
2. Allow to specify it when building `server`

First seems more reasonable to me. Since this will require some under-the-hood integration with GTO, I'd do this after #664 - which have the same decision to make.

fyi @omesser

aguschin:
close #665 

If you do:

```sh
git clone https://github.com/iterative/example-gto
cd example-gto
git checkout mlem
mlem --tb serve fastapi -m models/churn.pkl
```

You'll see model version on the last line:

image

TODO:
- [ ] add tests
- [ ] rename "version" to "mlem_version"
- [ ] review implementation - rough and needs to be refined

### DIFF
--- a/mlem/contrib/dvc.py
+++ b/mlem/contrib/dvc.py
@@ -21,6 +21,7 @@ from mlem.core.artifacts import (
     get_local_file_info,
 )
 from mlem.core.meta_io import get_fs
+from mlem.core.registry import ArtifactInRegistry
 
 BATCH_SIZE = 10**5
 
@@ -32,7 +33,7 @@ def find_dvc_repo_root(path: str):
     while True:
         if os.path.isdir(os.path.join(_path, ".dvc")):
             return _path
-        if _path == "/":
+        if _path == "/" or not _path:
             break
         _path = os.path.dirname(_path)
     raise NotDvcRepoError(f"Path {path} is not in dvc repo")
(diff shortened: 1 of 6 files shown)


mlem/contrib/dvc.py
```diff
@@ -118,3 +119,43 @@ def open(self) -> Iterator[IO]:
...
+def find_artifact_name_by_path(path):
+    if not os.path.abspath(path):
+        raise ValueError(f"Path {path} is not absolute")
+
+    from dvc.repo import Repo
+
+    root = find_dvc_repo_root(path)
+    relpath = os.path.relpath(path, root)
+    if relpath.endswith(".mlem"):
+        relpath = relpath[:-5]
```
omesser, line 133:
```suggestion
   relpath = relpath.rstrip(".mlem")
```

omesser (changes requested):

PR #694, opened 2023-07-13
Requirement version resolution fix

### LINKED ISSUE #688: Fixing version collection and allowing to override versions found
Is there a way to override the automatic detection of version, or a plan to add something like that? I had several issues with the automatic detection of dependencies, because of a mismatch between `MODULE.__version__`  and the version required for pip: 
- After installing `regex==2023.6.3`, `import regex  ; regex.__version__` gives `2.5.129` . But there is no `2.5.129` version of `regex` so the install fail
- After installing `opencv-python-headless=4.8.0.74`, `import cv2 ; cv2.__version__` gives `4.8.0` but there is no `4.8.0` version of `opencv-python-headless` (you need to give the 4th-level version).

I have a lockfile, and I use https://www.pantsbuild.org/ as a build system. So I can get the exact 3rd party dependencies. But I have not found a way to override the requirements inferred by MLEM.

Reported in [discord](https://discord.com/channels/485586884165107732/903647230655881226/1126406048882892960).

omesser:
Use `importlib.metadata`'s distribution to extract package versions
closes: https://github.com/iterative/mlem/issues/688


omesser:
I forgot about this for a few weeks but came back to this today and fixed the tests, with intention of merging this and releasing to fix  #688

omesser:
Seems that CI is broken (not related to changes here, see example in [this test PR](https://github.com/iterative/mlem/pull/701))
Will need to look at it separately and come back to this afterwards

PR #318, opened 2022-06-28
Allow access to secrets for external contributors

aguschin:
Using the changes from https://github.com/iterative/cml/pull/574

PR #695, opened 2023-07-13
CI: allow lint and short tests for everyone

omesser:
addresses: https://github.com/iterative/mlem/issues/330 (maybe partially, at least)
Allow basic checks and short ci tests without approval. need approvals for framework tests

### DIFF
--- a/.github/workflows/check-test-release.yml
+++ b/.github/workflows/check-test-release.yml
@@ -24,7 +24,6 @@ jobs:
     steps:
     - run: echo ✓
   check:
-    needs: authorize
     name: Check
     runs-on: ubuntu-latest
     steps:
@@ -49,8 +48,8 @@ jobs:
         key: pre-commit|${{ env.PYSHA }}|${{ hashFiles('.pre-commit-config.yaml') }}
     - run: pip install -U pre-commit tox
     - run: SKIP=pylint pre-commit run -a --show-diff-on-failure
-  test:
-    needs: authorize
+  test-short:
+    needs: check
     name: Test ${{ matrix.os }} with py${{ matrix.python }}
     runs-on: ${{ matrix.os }}
     strategy:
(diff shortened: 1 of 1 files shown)


aguschin (approved):

.github/workflows/check-test-release.yml
```diff
@@ -91,6 +90,40 @@ jobs:
...
+      run: pytest -k 'not heroku and not flyio'
+      env:
+        GITHUB_USERNAME: ${{ secrets.GH_USERNAME }}
+        GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
+        AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+        AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+        GITHUB_MATRIX_OS: ${{ matrix.os }}
+        GITHUB_MATRIX_PYTHON: ${{ matrix.python }}
+        BITBUCKET_USERNAME: ${{ secrets.BITBUCKET_USERNAME }}
+        BITBUCKET_PASSWORD: ${{ secrets.BITBUCKET_PASSWORD }}
```
aguschin, line 104: IIRC, initial idea was to allow external contributors to run tests that don't require using secrets #330 #318. But here you do exactly opposite, no?

aguschin (changes requested):
I'm not sure I understand this ^

.github/workflows/check-test-release.yml
aguschin: I'm trying to recall how we ended up with this `authorize` step, and IIRC, it was the case that CI/CD doesn't run for external contributors at all because GH doesn't allow them to use secrets we put to the repo. If that's true, we need to actually separate tests to two parts: one that requires secrets, and the other that doesn't.

omesser: I should've moved this back to draft, my bad sorry. I moved attention elsewhere but I'll catch up with you again before coming back to this.

PR #698, opened 2023-07-21
Larger runners for CI

omesser:
300Gb disk
Because there's a repeating out of disk space errors in CI


shcheklein (approved):

omesser:
Need to merge this to take effect and test

